PR #23, opened 2016-02-18
cascade can now check managed links in hero panel

mirabelrice:


app/assets/javascripts/widgets/hero_panel.js
```diff
@@ -0,0 +1,10 @@
+$(document).ready(function(){
+  var nav = $('.program-select');
```
jameskerr: Can you make this select a little more specifc?  Why don't you add an ID to the program select called `dropdown-select-links` then select that in the Javascript.

PR #69, opened 2016-09-01
[DEFECT] https causes pages with flickr/picasa feeds to not show thumbnails on www.

tatwell:
Trello card: https://trello.com/c/fnARAJmi/.

I'll need help deploying to dev and then production.

### DIFF
--- a/app/views/widgets/right_column/_picasa_gallery.html
+++ b/app/views/widgets/right_column/_picasa_gallery.html
@@ -2,7 +2,7 @@
 	<h3>Picasa Widget</h3>
     <ul>
     </ul>
-    <a class="more-link" href="javascript:;">
-              See Full Gallery <span class="arrow">»</span>
-            </a>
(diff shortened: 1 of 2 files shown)


app/views/widgets/right_column/_picasa_gallery.html
```diff
@@ -2,7 +2,7 @@
 	<h3>Picasa Widget</h3>
     <ul>
     </ul>
-    <a class="more-link" href="javascript:;">
-              See Full Gallery <span class="arrow">»</span>
-            </a>
+    <a class="more-link" href="https://get.google.com/albumarchive/105740368086394902806">
```
tatwell: The link we were building was broken. (I don't think Picasa even exists anymore. It's now Google Photos or Google Album or something.)

Any reason why we don't just hardcode the link here?

mirabelrice:
Looks good to me! Matt or I can walk you through deploying this today

PR #74, opened 2016-10-06
Fix flexslider

mirabelrice:


tatwell:
Well done. Did you recreate the changes in the unminified version of flexslider.js that BREI made? Can't tell in diff with new file. If so: 👍 !

mirabelrice:
yes, so the title field in the slider nav is outputting correctly now

PR #76, opened 2016-10-10
Refactor controllers as cascade content types.

tatwell:
Spike demonstrating new Rails pattern for for better mapping Cascade Assets development to Cascade.


app/models/configuration_set.rb
```diff
@@ -0,0 +1,32 @@
+#
+# Tableless model simulating Metadata Sets in Cascade: Administration > Configuration Sets
```
homeofmatt: Configuration Sets*

PR #81, opened 2016-10-17
Homepage hero fix

### LINKED ISSUE #80: Glitch in hero panel home page on Dev Cascade.
Most information here:
https://trello.com/c/2czI7Q2s/97-defect-glitch-in-hero-panel-home-page-on-dev-cascade

homeofmatt:
Closes #80


mirabelrice:
So we don't know how long this has been an issue for? or if something recently changed to cause it?

Looks good to me, assuming that index.js file is not needed anywhere else and the updated styles look fine.

app/assets/stylesheets/widgets/home_page/hero_panel.scss
```diff
@@ -21,7 +21,7 @@
   clear: both;
   position: relative;
   overflow: hidden;
-  min-height: 300px;
+  height: 300px;
```
tatwell: These min-height settings could explain it, no? They set a lower bound, but not an upper. So for images, or inner html, that exceeded that height the slide element might expand vertically.

Not sure why it started happening all the sudden. But perhaps something got tweaked in the way the flexslider manhandles the DOM.

homeofmatt: I mean I guess they explain it in a sense. Those min-height settings were certainly the root of the problem there, but I'm still a little miffed as to why they weren't causing issues before.

PR #84, opened 2016-10-26
Homepage link color and sitewide font fixes

waidelacruz:
- Homepage link color changes 
- Proxima-Nova-Soft change to Proxima-Nova --- per David and Jeff
- Correct font-family value syntax
- Correct grey hex value and font weight

### DIFF
--- a/app/assets/stylesheets/base/defaults.scss
+++ b/app/assets/stylesheets/base/defaults.scss
@@ -20,7 +20,7 @@ html {
   -ms-text-size-adjust: 100%;
   background-image: image-url('window-pattern.gif');
   background-color: $cu-light-grey;
-  font-family: "proxima-nova-soft", "proxima-nova-soft-n4", sans-serif;
+  font-family: "proxima-nova", verdana, sans-serif;
   overflow-x: hidden;
   -webkit-font-smoothing: antialiased;
   &.cu-companion-bar {
@@ -34,7 +34,7 @@ body {
   margin: 0;
   padding: 0;
   color: $cu-font-color;
-  font-family: "proxima-nova-soft", "proxima-nova-soft-n4", sans-serif;
+  font-family: "proxima-nova", verdana, sans-serif;
   &.personnel .main {
 	  margin-left: 0;
 	}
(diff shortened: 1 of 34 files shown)


app/assets/stylesheets/base/defaults.scss
```diff
@@ -108,7 +108,7 @@ a {
     color: $white;
     display: inline-block;
     text-transform: uppercase;
-    font-family: 'proxima-nova', sans-serif;
+    font-family: "proxima-nova", verdana, sans-serif;
```
tatwell: Any reason these can't be assigned as a SASS variable, like the colors?

tatwell:
A few suggestions that can be taken or leaven. Otherwise 👍 .

app/assets/stylesheets/base/defaults.scss
tatwell: I'll leave that to you. Whatever is simpler.

app/assets/stylesheets/widgets/home_page/blog_stories.scss
```diff
@@ -242,7 +242,7 @@
         display: inline-block;
         padding: 5px 10px 5px 20px;
         /*background:#f7f6f0;*/
-        /*border-left:5px solid #a82439;*/
+        /*border-left:5px solid #98002e;;*/
```
waidelacruz: @tatwell I usually don't remove any comments in case anyone put them there for a reason. Should we make it standard practice to remove dead code when we see them?

tatwell: Yes. If there's a reason for not removing them, they should include a comment explaining why. Otherwise they just tend to cause confusion.

app/assets/stylesheets/base/defaults.scss
waidelacruz: @tatwell I assigned instances of proxima (body) and futura (headers) font famlies to sass variable as suggested. Makes sense to do them and I think we should switch to variables moving forward.

tatwell: @waidelacruz Did you push your changes? I don't see the new font-family variable.

tatwell:
I like the new SASS variables. Unconditional 👍 .

mirabelrice:
👍

PR #85, opened 2016-10-26
Stories feed widget

mirabelrice:


lib/assets/javascripts/chapman_stories_feed.js
```diff
@@ -0,0 +1,191 @@
...
+  // passed in so we can ensure the value of it is truly undefined. In ES5, undefined
+  // can no longer be modified.
+
+  // window and document are passed through as local variables rather than global
+  // as this (slightly) quickens the resolution process and can be more efficiently
+  // minified (especially when both are regularly referenced in your plugin).
+
+  // Create the defaults once
+  var defaults = {
+    baseUrl: 'http://social.localchapman.edu:3000/',
```
mirabelrice: I will change this back to social.chapman.edu once inside is deployed

lib/assets/javascripts/chapman_stories_feed.js
```diff
@@ -0,0 +1,191 @@
...
+      return $outerContainer;
+    },
+
+    buildTopStoryElement: function() {
+      var classes = 'feed-column top-story';
+      var loadingHtml = '<p>Loading top story...</p>'
+      var $topStoryElement = $('<div />').addClass(classes).html(loadingHtml);
+      
+      //if top story post is not specified, default to the top story from the specified blog feed
+      var params = this.topStoryUrl ? { url: this.topStoryUrl } : { source: this.feedUrl };
```
mirabelrice: Here is the logic for the top story default

tatwell: Send both and let the backend (Inside) figure it out. That way, if we decide we want to change the logic or output, we can do it there and not worry about coordinating with this.

lib/assets/javascripts/chapman_stories_feed.js
```diff
@@ -0,0 +1,191 @@
...
+      this.buildOutWidget();
+    },
+
+    //this is a temporary fix to remove trailing slash from blog site urls(that is how they are stored on inside db)
+    //Ideally, inside query should be able to match urls with or without trailing slash
+    parseFeedUrl: function() {
+      var parsedUrl = this.$element.data('chapmanStoriesFeedUrl') || this.settings.dataFeedUrl;
+      parsedUrl = parsedUrl.replace(/\/?$/, '');
+      return parsedUrl;
+    },
```
tatwell: This isn't used, is it? I think this is a vestige of previous version. If so, please remove.

mirabelrice: Yes this is being used until the query can be fixed in ElasticSearch

lib/assets/javascripts/chapman_stories_feed.js
mirabelrice: okay, fixed

app/assets/stylesheets/widgets/single_column/chapman_stories_feed.scss
```diff
@@ -0,0 +1,151 @@
+.chapman-stories-feed-widget {
+	padding: $one-column-widget-spacing 0;
+  text-align: center;
```
mirabelrice: Yes my editor is using 2 spaces so i'm not sure why the formatting is off here.

tatwell:
👍

PR #100, opened 2016-11-16
Pre-Mortem: New cascade mastheads (NO MERGE)

tatwell:
This PR is for discussion/review only.

DO NOT MERGE. A new PR will be issued when ready.


app/data_definitions/from_cascade/two_column.xml
```diff
@@ -0,0 +1,731 @@
...
+            <radio-item value="No"/>
+        </text>
+        <text type="radiobutton" identifier="showImage" label="Custom Image" default="No" help-text="for OLD Masthead. Indicates to use Image below instead of default image">
+            <radio-item value="Yes"/>
+            <radio-item value="No"/>
+        </text>
+        <asset type="file" identifier="image" label="Image (1024x260)" help-text="for OLD Masthead. Large image to replace small default image"/>
+        <text identifier="altText" label="Image Alt-Text" help-text="Alternate text to describe image (required)"/>
+        <text identifier="description" label="Description" maxlength="150" help-text="for OLD Masthead. Visible caption under masthead image"/>
+    </group>
```
tatwell: End `masthead` group.

tatwell: Here's the full old masthead XSLT block:

```
<xsl:when test="showMasthead='Yes' and showImage='Yes'">
    <div class="masthead-old">
        <xsl:if test="link/label/text() != ''">
            <xsl:attribute name="class">masthead withLink</xsl:attribute>
        </xsl:if>
        <xsl:if test="showImage='Yes'">
            <xsl:if test="not(image/path='/')">
                <xsl:if test="not(altText='')">
                    <img alt="{altText}" src="{image/path}" width="1024"/>
                </xsl:if>
                <xsl:if test="altText=''">
                    <img alt="[photo description needed in old masthead]" src="{image/path}" width="1024"/>
                </xsl:if>
            </xsl:if>
        </xsl:if>
        <div>
            <xsl:value-of select="description"/>
            <xsl:if test="link/label/text() != ''">
                <a class="more">
                    <xsl:choose>
                        <xsl:when test="not(link/fileLink/path='/')">
                            <xsl:attribute name="href">
                                <xsl:value-of select="link/fileLink/path"/>
                            </xsl:attribute>
                        </xsl:when>
                        <xsl:when test="not(link/internalLink/path='/')">
                            <xsl:attribute name="href">
                                <xsl:value-of select="link/internalLink/path"/>
                            </xsl:attribute>
                        </xsl:when>
                        <xsl:otherwise>
                            <xsl:attribute name="href">
                                <xsl:value-of select="link/link"/>
                            </xsl:attribute>
                        </xsl:otherwise>
                    </xsl:choose>
                    <xsl:value-of select="link/label"/>
                </a>
            </xsl:if>
        </div>
    </div>
</xsl:when>
```

The same field is used for the more link in last block. 

I suspect that block doesn't exist and these tests always return false. Is that the case?

mandythomas: I don't see a link/label field at all in the data definition for the "old" masthead. There may have been one ages ago though I actually don't remember them ever allowing a link in them. I believe that part of the format is unnecessary. I confirmed with Ross and he doesn't remember the old mastheads ever allowing a link from the photo or description or anything either

tatwell:
Pre-mortem review. Closing without merge.

PR #103, opened 2016-11-21
As a web coordinator in Cascade, I want to be able to choose Wai's new mastheads in Cascade.

tatwell:
https://trello.com/c/nyCOSWbo/74-as-a-web-coordinator-in-cascade-i-want-to-be-able-to-choose-wai-s-new-mastheads-in-cascade#


app/assets/stylesheets/regions/mastheads/slider-masthead.scss
```diff
@@ -1 +1,70 @@
...
+          background-color: rgba(56,56,56,0.90);
+        }
+      }
+
+      &.header {
+        width: 350px;
+        color: $white;
+
+        div.aligned {
+          position: absolute;
```
mirabelrice: Is there a reason for the `div` selectors? and is it necessary to make the selectors this specific? I think we want to avoid nesting more than 2 or 3 levels deep if possible.

mirabelrice: Yes, however unnecessarily specific styling can lead to a "css selector nightmare" as it makes code unmodifiable as stated in the sass style guide http://thesassway.com/beginner/the-inception-rule

homeofmatt: Hmm... I mean Sass is classy and all, but I feel like it can be easy to let nesting get out of hand with it.
A CSS selector that looks like `div.slider.version-201611 div.grid .column.image div.photo-by` or `div.slider.version-201611 div.grid .column.header div.aligned h2` to name a couple from this file just makes me shudder.

app/controllers/content_types/modular_controller.rb
```diff
@@ -87,6 +87,120 @@ def one_column
...
+      # Set dynamic values.
+      if params[:masthead] == 'branded-old'
+        @data_definition.set_value(:masthead_type, 'Branded Masthead')
+      elsif params[:masthead] == 'slider-old'
+        @data_definition.set_value(:masthead_type, 'Slider')
+      elsif params[:masthead] == 'slider'
+        @data_definition.set_value(:masthead_type, 'Slider - New')
+      else
+        @data_definition.set_value(:masthead_type, 'Branded - New')
+      end
```
homeofmatt: Why not make a method that takes the `[:masthead]` parameter name and dynamically assigns the `:masthead_type` value to the data definition instance? I feel like that's probably feasible here and we would just have to ensure our naming is 1-1 with these files.
In fact, that would probably help with organization. Having a param `slider` not be equal to the `masthead_type: 'Slider'` is confusing to me.

app/controllers/content_types/modular_controller.rb
```diff
@@ -87,6 +87,120 @@ def one_column
...
+      @data_definition = DataDefinitions::ThreeColumn.default
+
+      # Set dynamic values.
+      if params[:masthead] == 'branded-old'
+        @data_definition.set_value(:masthead_type, 'Branded Masthead')
+      elsif params[:masthead] == 'slider'
+        @data_definition.set_value(:masthead_type, 'Slider - New')
+      else
+        @data_definition.set_value(:masthead_type, 'Branded - New')
+      end
```
homeofmatt: Same as previous comment on consolidating conditionals to one abstracted method.

app/data_definitions/base.rb
```diff
@@ -27,12 +44,33 @@ def get_child(field)
...
+        node = at_xpath(xpath)
+        raise "Node not found for xpath #{xpath}." unless node
+        node.content
+      end
     end
-    # rubocop:enable Rails/OutputSafety
   end
+
+  # Monkey-patch it now!
+  Nokogiri::XML::Element.include DataDefinitions::NokogiriExtensions::Element
```
homeofmatt: Not saying anything is wrong, I just see a comment saying that something is monkey patched. What's the reasoning?

app/data_definitions/three_column.rb
```diff
@@ -0,0 +1,73 @@
...
+          ]
+        }
+      }
+    }.freeze
+
+    # Class Methods
+    def self.default
+      three_column = DataDefinitions::ThreeColumn.new
+      three_column.set_defaults
+      three_column
```
homeofmatt: This could be a one line method if we did:
```
def self.default
   DataDefinitions::ThreeColumn.new.set_defaults
end
```

app/data_definitions/two_column.rb
```diff
@@ -0,0 +1,113 @@
...
+        }
+      }
+    }.freeze
+
+    # Class Methods
+    def self.default
+      two_column = DataDefinitions::TwoColumn.new
+      two_column.set_defaults
+      two_column
+    end
```
homeofmatt: Same comment as before:
```
def self.default
   DataDefinitions::TwoColumn.new.set_defaults
end
```

app/models/configuration_set.rb
```diff
@@ -31,6 +31,22 @@ def self.one_column(options={})
...
+  end
+
+  def self.three_column(options={})
+    default_template_path = '_cascade/templates/modular/three_column.html'
+    configuration_set = ConfigurationSet.new(name: '3 Column')
+    configuration_set.set_defaults
+    configuration_set.template = options.fetch(:template, default_template_path)
+    configuration_set
+  end
+
```
homeofmatt: Similar to my last two comments about refactoring those `self.default` class methods. I just personally like avoiding using a whole line to return an object like here with `configuration_set`.  I feel like it looks a little messy and it doesn't strike me as Ruby best practice if it's avoidable. I really try to take advantage of the Ruby behavior of returning whatever is produced by the last line of the method.

Not sure if there's such a simple solution here as there was on the last two methods just from precursory glance, but that's my 2 cents on Ruby philosophy.

mirabelrice:
Overall i'm okay with this 👍  Im still a bit concerned with transitioning cascade data into this app and  unsure of the what the workflow is. Maybe this will be cleared up over time or can be addressed in the README.

app/views/_cascade/formats/level/_masthead.erb
```diff
@@ -0,0 +1,240 @@
...
+                if slideInternalLinkPath != '/'
+                  slideCaptionHref = slideInternalLinkPath
+                elsif slideInternalLinkPath == '/' && slideLinkUrl != ''
+                  slideCaptionHref = slideLinkUrl
+                elsif slideInternalLinkPath == '/' && slideLinkUrl == '' \
+                  && slideFileLinkPath != '/'
+                  slideCaptionHref = slideFileLinkPath
+                else
+                  slideCaptionHref = ''
+                end
```
homeofmatt: Looking at this I totally understand why this is written the way it is if this is how our Velocity looks on Cascade and this is just replicating that behavior, but phew. Two-liner conditionals with 3 conditions stiiiiiiiinks 💩

app/views/_cascade/formats/modular/_page_wrapper_open.erb
```diff
@@ -0,0 +1,19 @@
...
+
+currentPagePath.split('/').each do |name|
+  if currentPagePath.split('/').first == name
+    baseClass = name
+    sectionClasses = baseClass
+  elsif currentPagePath.split('/').last != name
+    baseClass = "#{baseClass}__#{name}"
+    sectionClasses = "#{sectionClasses} #{baseClass}"
+  end
+end
```
homeofmatt: Not sure what's going on in the big picture here, but I'm just curious about what happens if `currentPagePath.split('/').last == name`

I just don't often see a conditional block with a `==` followed by an elseif with `!=`

app/assets/stylesheets/regions/mastheads/slider-masthead.scss
tatwell: @mirabelrice I agree blindly nesting like the article describes is bad (see line 5 of the Zen: `Flat is better than nested.`), but that's not what's really happening here. 

Notice, there's no empty rules. Rather, I'm targeting a specific case, e.g. the new slider master, to the exclusion of existing slightly variant versions. I've inherited a deeply nested structure here with the flex-slider. The SASS rules target specific pieces of it.

But I'm always open to cleaner alternative, provided they don't hide complexity or produce conflicts.

tatwell: @homeofmatt I agree that complexity/nesting should be avoided, but that comes from the HTML, so I don't think it can be avoided here. I'd prefer the Sass acknowledge the complexity that's beyond it control rather than hide it.

app/controllers/content_types/modular_controller.rb
tatwell: I agree in future something may need to be done. For now, I think this is manageable and easier to comprehend.

The naming mismatch is brought over from Cascade. We need to map URL param values (with their URL-encoded format) to the Cascade labels we used.

I'll add a comment clarifying this.

homeofmatt: I don't mind taking it on for today and tomorrow to refactor these conditionals

homeofmatt: Well alright then, you know more about the situation. I just really really don't like behavior like this. It would already take a bit of time to refactor and make sure nothing broke, I just can't imagine what it would be like if we were refactoring these conditionals with 10/15/20 masthead options.

app/data_definitions/three_column.rb
tatwell: It could but I actually prefer multiple lines here for readability and clarity.

The one-liner works if `set_defaults` returns `self`, but I don't think it does at moment. And it's not something I would necessarily expect it to do. I guess this again reveals my un-railsy explicitness bias.

There's this consideration, too: there can be other class methods like this for alternative data profiles, e.g. `DataDefinitons::ThreeColumn.admission_undergraduate_index`. In those, you would probably be explicitly assigning values. This also provides a clearer template for those.

app/views/_cascade/formats/level/_masthead.erb
```diff
@@ -0,0 +1,240 @@
...
+                slideQuoteAuthor = slide.select_single_node_value('quote-author')
+                slideCaptionAlign = slide.select_single_node_value('align')
+
+                # Set slide href.
+                if slideInternalLinkPath != '/'
+                  slideCaptionHref = slideInternalLinkPath
+                elsif slideInternalLinkPath == '/' && slideLinkUrl != ''
+                  slideCaptionHref = slideLinkUrl
+                elsif slideInternalLinkPath == '/' && slideLinkUrl == '' \
+                  && slideFileLinkPath != '/'
```
tatwell: Yep. It's translated straight from the Cascadese. (See the javascript above.)

We need to get stricter with code like this, especially where vendors are involved. Problem is VTL code is outside repo and PRs.

app/data_definitions/three_column.rb
homeofmatt: Ah yea makes sense. I more just noticed the one liner to return the object. I explain my feelings on that behavior a couple of comments down.

app/views/_cascade/formats/modular/_page_wrapper_open.erb
tatwell: This tries to replicate some VTL code in Cascade. So I think it follow their logic. In any event, it works for our purposes here -- testing themes -- and doesn't get folded back into anything functional. I treat it more as a test helper.

I'll add a comment to clarify that.

tatwell:
> Im still a bit concerned with transitioning cascade data into this app and unsure of the what the workflow is.

Yes, I'm figuring this out as I go. I'm satisfied with the result so far and I think it did help immensely with this story.

The next time we work on a Cascade story would probably be a good time to review the workflow. It's pretty clear in my head right now and I've tried to be as intuitive and clear with the code I've written. But that will be a good time to put my assumptions to the test.

homeofmatt:
:+1:

tatwell:
Product owner approved. I neglected mobile design so that will be address in 2 new user stories (1 for each design).

PR #109, opened 2016-11-30
made slider masthead mobile friendly

mirabelrice:


app/assets/javascripts/cascade/level/rotator.js
```diff
@@ -31,7 +31,7 @@ $(function () {
         var beforeSlideChange = function(slider) {
             if ( !isSliderMasthead ) { return; }
 
-            $header.fadeOut();
```
mirabelrice: Switched this to fadeTo because fadeOut/fadeIn makes the header jump when the height is not fixed

tatwell: I wasn't sure what this difference was. [This StackOverflow comment](http://stackoverflow.com/questions/3311758/fadein-vs-fadeout-vs-fadeto#comment41672608_3311790) explains it: `fadeTo` doesn't tweak the `display` property.

app/assets/stylesheets/regions/mastheads/slider-masthead.scss
```diff
@@ -29,44 +51,69 @@ div.slider.version-201611 {
...
-            background-color: white;
-            border: none;
+              display: block;
+              height: 2px;
+              width: 35%;
+              margin:0 auto;
+              color: white;
+              background-color: white;
+              border: none;
+            }
```
tatwell: This looks to me like `hr` line will only show on medium screens and not larger ones (as I would expect). Am I reading this correctly?

Checking the console in Chrome, I see I'm not. `$medium-screen` means 780px or larger.

app/data_definitions/two_column.rb
```diff
@@ -77,11 +77,11 @@ class TwoColumn < DataDefinitions::Base
 
       v201611: {
         slider: {
-          header: 'Campus Life \n at \n Chapman University',
+          header: 'Campus Life at Chapman University',
 
           slides: [
             {
-              subTitle: '',
+              subTitle: 'Subtitle Here',
```
tatwell: Changes look good. Only these last two here concern me. Restore these and if there are causing issues, they should be worked out with product owner.

app/assets/javascripts/cascade/level/rotator.js
mirabelrice: You are still seeing the jump? Im not seeing it with fadeTo.

tatwell: By jump, I mean the block gets larger. I don't know what else it could do not being fixed height. What did you mean originally by jump?

mirabelrice: Oh I see what you're saying. By jump I meant that the red header block would completely disappear and reappear on slide change. The height jump that is currently happening is because some slides have more text and the height needs to adjust. Wai said that is fine as there is not really a way around that.

app/data_definitions/two_column.rb
mirabelrice: got it. fixed.

app/assets/javascripts/cascade/level/rotator.js
tatwell: Ok. That makes sense. Yes, this is better. I would add a comment explaining this. Future dev might be tempted to add back fadeOut/fadeIn thinking it does the same thing (which it almost does).

PR #114, opened 2016-12-08
omninav does not cover target when anchor links are clicked

mirabelrice:


app/assets/javascripts/widgets/wysiwyg.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function(){
+  var anchorLinks = $('h3.anchorLinks a, a[href="#top"]');
+  var navHeight = $("#cu_nav").outerHeight();
+
+  /* offset omninav when interal anchor links scroll to target */
+  anchorLinks.on('click', function(e){
+    var target = $('[name=' + this.hash.slice(1) + ']');
```
tatwell: Is target an attribute of the `e` event object? I believe `e` usually has a `target` attribute. If it is one and the same, I would much prefer to use that.

What is the `this` value in this case? And what is `this.hash`?

app/assets/javascripts/widgets/wysiwyg.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function(){
+  var anchorLinks = $('h3.anchorLinks a, a[href="#top"]');
+  var navHeight = $("#cu_nav").outerHeight();
```
tatwell: There is a convention (which I like and use myself) of starting any var that is a jQuery element object or collection with `$`. So that this would be `$navHeight`. Should we adopt that as part of our style guide?

app/assets/javascripts/widgets/wysiwyg.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function(){
+  var anchorLinks = $('h3.anchorLinks a, a[href="#top"]');
```
homeofmatt: Agreed ^
We should add it to our style guide

app/assets/javascripts/widgets/wysiwyg.js
mirabelrice: I just named the variable target because it is the scroll target. `this` is referring to the anchor element that was clicked. and `this.hash` returns the anchor part of a url.

tatwell: I want to believe there is a way to access the `target` var as a property of the `e` object rather than having to run a jQuery selector to get it.

But your explanation does make what's going on here a lot clearer. If you keep this, can you add that info to comment?

mirabelrice: The `event` object is referring to the link that was clicked. This selector is finding the page location to scroll to, which is identified by the name attribute.

tatwell: You're right. Ignore my previous comment.

PR #115, opened 2016-12-08
vimeo video callout

homeofmatt:


tatwell:
👍 ... so far.

I assume there would be additional changes required if we needed, for instance, to extract and format a Vimeo link from the user input.

homeofmatt:
Yea there would be in velocity with formats.

PR #122, opened 2016-12-13
Background image update

waidelacruz:
Update the background window pattern on www to include the gradient and the use of white lines per Wai's latest design comp.

https://trello.com/c/nZHq9VYr


tatwell:
👍

PR #130, opened 2017-01-05
Add modular formats

homeofmatt:


app/models/configuration_set.rb
```diff
@@ -32,18 +32,20 @@ def self.one_column(options={})
...
   end
 
   def self.three_column(options={})
-    default_template_path = '_cascade/templates/modular/three_column.html'
+    # default_template_path = '_cascade/templates/modular/three_column.html'
     configuration_set = ConfigurationSet.new(name: '3 Column')
     configuration_set.set_defaults
-    configuration_set.template = options.fetch(:template, default_template_path)
+    configuration_set.template = options.fetch(:template, 
+                                                '_cascade/templates/modular/three_column.html')
```
tatwell: Any reason for these changes? This is what I originally had it. I changed it to make it a bit more templaty, I guess.

If this is better practice, I'm ok with it. I'd just remove the commented out line.

homeofmatt: Oh I just changed them to be the same as the ad_landing method. I don't really care either way

tatwell: Meh. Just leave it then.

PR #133, opened 2017-01-12
[NO MERGE] Reproduce primary content block as Cascade format in one column modular template. 

tatwell:
**Note: This branch is not ready to merge. If it has not been merged by 2017-06-30, I recommend it be deleted.**

The goal here is to reproduce as closely as possible the syntax and logic of the Cascade format for the primary content region in the one column modular template. That way, we could develop these scripts in Cascade Assets, put them under source control, and then move them over to Cascade with as little translation as possible.

It introduces a new `Cascade` library under `lib`.

I tried to do this by bending Rails / Ruby syntax, but it broke down once I got to Cascade Velocity macros. If you have ideas for working around this, you are welcome to take it up. I'd recommend discussing it with me first.

An alternative approach would be to build an actual interpreter for VTL, kinda  like Slim:

https://github.com/slim-template/slim

But that would be well beyond the scope of this project right now.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,8 @@ class ApplicationController < ActionController::Base
 
   def cascade_format(format_path)
     # Make data definition a local within partial as variable current_page
-    render_to_string(partial: format_path, locals: {current_page: @data_definition})
+    render_to_string(partial: format_path, locals: {current_page: @data_definition,
+                                                    contentRoot: @data_definition.document})
   end
 
   def render_velocity(format_path, data)
--- a/app/controllers/content_types/modular_controller.rb
+++ b/app/controllers/content_types/modular_controller.rb
@@ -75,11 +75,11 @@ module ContentTypes
         'OG_TAGS' => '',
         'PAGE WRAPPER CLOSE' => '',
         'PAGE WRAPPER OPEN' => '',
+        'PRIMARY CONTENT' => cascade_block('_cascade/formats/modular/one_column_primary_content'),
 
         # TODO: convert these to cascade_format action.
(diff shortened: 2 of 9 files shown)


app/controllers/content_types/modular_controller.rb
```diff
@@ -61,6 +61,17 @@ def one_column
...
+      # Manually set primary content.
+      widget = @data_definition.document.at_xpath('//group[@identifier="widget"]')
+      widget.at_xpath('text[@identifier="widgetType"]').content = 'Image with Text'
+
+      # TEST
+      currentPage = Cascade::XPathTool.selectSingleNode(@data_definition.document, "//system-page[@current]")
+      systemData = currentPage.getChild('system-data-structure')
+      primaryContent = systemData.getChild('primaryContent')
+      widgets = Cascade::XPathTool.selectNodes(primaryContent, "widget")
+      p currentPage.name, systemData.name, widgets.length
```
tatwell: This is debugging cruft.

app/controllers/application_controller.rb
```diff
@@ -11,7 +11,8 @@ def cascade_block(block_path)
 
   def cascade_format(format_path)
     # Make data definition a local within partial as variable current_page
-    render_to_string(partial: format_path, locals: {current_page: @data_definition})
+    render_to_string(partial: format_path, locals: {current_page: @data_definition,
+                                                    contentRoot: @data_definition.document})
```
tatwell: `current_page` should be dropped here, but it's used by some of the other `format` partials.

tatwell:
Wasn't merged before 2017-06-30 so I'm going to close it without merging and delete as recommended.

PR #144, opened 2017-02-07
removed flash download link from noscript tags

mirabelrice:


tatwell:
👍  

Obviously updating 10 files is a bit of a 0_o.

PR #145, opened 2017-02-07
Development

mirabelrice:
added aria-label attribute to search drop down


tatwell:
👍  You can merge this without review.

Issue #160, opened 2017-03-03
events listing in News Events widget of 2/3 Column Modular templates is not loading


mandythomas:
The events listing within the News/Events widget in 2/3 Column Modular templates is not working on most pages. One where it works is http://www.chapman.edu/dodge/student-life/index.aspx. One where it doesn't is http://www.chapman.edu/business/argyros-alumni/index.aspx. 

It uses rss feeds from events.chapman.edu and in particular the category field for the date of the event. The category and pubdate fields both contain the event date. Category is being parsed in the news-events.js javascript to pull out parts of the date. However recently additional category nodes are appearing in the rss feed for any given event and they contain info about buildings, not a date, so the function is failing. See https://events.chapman.edu/events.rss?group_id=11%2C73%2C29%2C92%2C31%2C163%2C10

Need to figure out why extra category nodes are in the rss now, OR use the pubdate instead of category field in the javascripts\cascade\level\news-events.js function, however there was a note in javascript function saying "pubdate sometimes contained original but not current event date; use category field instead". Need to verify if that's still true or safe to use. Also note the dates are formatted differently in pubdate vs category nodes.

tatwell:
Trello card: https://trello.com/c/tej9kISv/

- Examples
  - Failing: https://www.chapman.edu/events/homecoming-weekend/index.aspx
  - Working: https://www.chapman.edu/dodge/student-life/index.aspx

Also, loading image broken and needs to be fixed:

    <img alt="Loading events list..." class="loading" src="../_files/level/img/ajax-loader.gif">

tatwell:
Events widget source:

- /_cascade/formats/modular/widgets/Featured News Events Left Col

tatwell:
I think I've identified the source of the issue. Not sure why yet it's happening.

I've reproduced the issue here:
- http://localhost:3000/modular/two_column

When I load that page, I see the following error in the console:
```
Uncaught TypeError: rssitem.category.split is not a function
```

I added a debug statement to report the data that was returned from the Yahoo API feed:
```
//query.yahooapis.com/v1/public/yql?q=select%20*%20from%20rss(3)%20where%20url%3D'http://www.chapman.edu/getFeed.ashx?name=events'&format=json&diagnostics=true&callback=?
```

The data includes 3 `data.query.results.item` objects. Here is what it reports for each `category` element:

- `["2017/03/06 (Mon)", "WALLACE ALL FAITHS CHAPEL"]` (Array)
[...]

PR #161, opened 2017-03-06
On mobile and on desktop, sometimes the events feed on WWW does not work.

tatwell:
Fixes Issue #160.

Trello card: https://trello.com/c/tej9kISv/261-defect-on-mobile-and-on-desktop-sometimes-the-events-feed-on-www-does-not-work


app/assets/javascripts/cascade/level/news-events.js
```diff
@@ -1,407 +1,450 @@
 $(function () {
```
tatwell: Fix is in this file. I also cleaned up formatting so diff here is useless. Find relevant changes here:

- https://github.com/chapmanu/cascade-assets/pull/161/commits/072df4de9071a6069ab9592f4a529f7a42708930

mandythomas:
@tatwell I reviewed one of your test pages, and also copied my own and changed feed to ASBE which wasn't working before, and also put the widget in primary content and not just the one in the left column. Both worked correctly now. I approve pull request

Issue #162, opened 2017-03-07
Search results are unreadable.


tatwell:
It looks like a background-color setting was lost for search results widget.

To reproduce, simply go to production home page and do a search:

- https://www.chapman.edu/

Ross reports he started seeing it around 4pm today. Prior to that, the results black has a white background.

tatwell:
This is the closest selector for the affected block: `#cu_search_results_ui`

Most likely problem was caused by recent change in one of these files:

- app/assets/stylesheets/components/cu-search-results.scss
- app/assets/stylesheets/omni_nav/modules/omni-nav.scss

tatwell:
Well, [cu-search-results.scss](https://github.com/chapmanu/cascade-assets/blob/development/app/assets/stylesheets/components/cu-search-results.scss) has been completely commented out since last June.

TODO: Remove this file.

That leaves [omni-nav.scss](https://github.com/chapmanu/cascade-assets/blob/development/app/assets/stylesheets/omni_nav/modules/omni-nav.scss), but that file has been updated since Feb 8.

Which means perhaps a change in a global sass variables file?

tatwell:
I'm seeing this error in the console when I submit a search in production:

     Refused to display 'https://cse.google.com/cse_v2/ads?q=hello&r=m&cx=015856566681218627934%3A2n…sd13sv13lhsl1sr1-&jsv=11710&rurl=https%3A%2F%2Fwww.chapman.edu%2F#master-1' in a frame because it set 'X-Frame-Options' to 'SAMEORIGIN'.

Perhaps it has something to do with the issue.

tatwell:
Trello card: https://trello.com/c/cBHSE5LT/274-urgent-in-www-site-search-results-are-misstyled-and-hard-to-read

tatwell:
Pull up sample home page in dev server:

http://localhost:3000/home_page/sample

Results here have the expected white background. It applied to this element:

    div.gsc-control-cse.gsc-control-cse-en

The following style is applied:

However, the stylesheet actually comes from Google:

- https://www.google.com/cse/static/style/look/default.css

So that appears to be the issue. The question is: why did production and dev stop loading this stylesheet?

tatwell:
Released to production.

PR #163, opened 2017-03-07
In WWW site, search results are misstyled and hard to read.

tatwell:
Fixes Issue #162.

Trello Card: https://trello.com/c/cBHSE5LT


tatwell:
@mandythomas Interesting. I guess it was a temporary glitch on Google's part where they stopped serving up the stylesheet. This fix should insure that results are readable even if this happens in future.

PR #169, opened 2017-04-14
created brochure masthead

mirabelrice:


app/assets/javascripts/widgets/brochure_masthead.js
```diff
@@ -0,0 +1,6 @@
+$(document).ready(function() {
+  $(".offcanvas-menu-trigger a").on("click", function(){
+    $("#js-cu-off-canvas-nav-trigger").trigger("click");
+    return false;
```
tatwell: This code looks odd to me. Shouldn't this be covered by existing OmniNav functionality?

app/data_definitions/from_cascade/one_column.xml
```diff
@@ -4,7 +4,13 @@
             <dropdown-item value="None"/>
             <dropdown-item value="Side panel hero" show-fields="masthead/sidePanelHero"/>
         </text>
-        <group identifier="sidePanelHero" label="Side panel hero">
+        <!-- (April 2017) Product owners request that only web coordinators be able to select new Brochure Masthead (temporarily) -->
+        <!-- This an ugly work around for selecting restricted mastheads.It can be removed when the type is available to all content creators -->
```
tatwell: Excellent comment.

app/assets/javascripts/widgets/brochure_masthead.js
mirabelrice: This is for the menu button on the brochure masthead. It triggers the off canvas-nav to open. I can move this Event handler into the omninav js, but since it is only for brochure pages I thought it made more sense here. What do you think?

mirabelrice: Also maybe it would make more sense for the js file to be named`brochure.js` as there will probably be more js needed for the rest of the page layout.

tatwell: Ok. I looked again at the design and it makes more sense. Maybe add a comment noting this is for the VIEW MENU button in masthead. Also, I like `brochure.js` suggestion.

PR #171, opened 2017-04-19
Left-column Shared-content Dept Contacts widget

mandythomas:
This pull request is for new left-column shared-content widget for Department Contacts. Has social icons as well which use inline SVG (was going to use icomoon-ultimate fonts but Wai changed direction and doing SVG paths now). Also created 3 samples of variations of widget to show in sample pages. Added new JS for the view more/less expander. Expander color changes per school-themes so also updated those css files.


waidelacruz:
Thanks @mandythomas ! I'll  clean up the icomoon-ultimate references much later when I convert the OmniNav fonts to svg.

PR #172, opened 2017-04-21
updated events widget design

mirabelrice:


lib/assets/javascripts/chapman_events_feed.js
```diff
@@ -90,14 +102,47 @@
...
+    var date_arr = date.split(",");
+    var date_html = [
+      '<div class="event-date"><div class="date-wrapper">',
+        '<div class="left date-column"><p>'+ date_arr[0] +'</p></div>',
+        '<div class="right date-column"><p>'+ date_arr[1] +'</p><p>'+ date_arr[2] +'</p></div>',
+      '</div></div>'
+    ];
+
+    return date_html.join('\n');
+  };
```
tatwell: I think this is ok. Better than straight string concatenation. The alternative using jQuery might be something like:

```
$event_date = $('<div />').addClass('event-date');
$date_wrapper = $('<div />').addClass('date-wrapper');
// create other DOM elements.
$date_wrapper.append($left_date_col).append($right_date_col);
$event_date.append($date_wrapper);
```

In this case, your code is probably more readable and concise.

We should add this topic to our best practices guide.

PR #182, opened 2017-05-04
Build of accessibility template and JS changes

sharissis-brei:


subprojects/degrees-and-programs/source/cascade/formats/Degrees-programs/PrimaryContent.vm
```diff
@@ -272,13 +269,15 @@
 
 										#foreach ( $motivation in $motivations )
 
+											#set ( $rand = $_MathTool.floor( $_MathTool.mul( $_MathTool.getRandom(), 10000 ) ) )
```
tatwell: Please add comment explaining random values here and with interests below.

subprojects/degrees-and-programs/source/cascade/formats/Degrees-programs/data/degrees.vm
```diff
@@ -17,7 +17,7 @@ ${output.replaceAll( $trimSpacingRegExp, "" )}
     ## Array Macro
     #macro ( generateArray $arr, $items )
         #foreach ( $item in $items )
-			#if ( $item != "" )
+    		#if ( $item != "" )
```
tatwell: Looks like formatting is inconsistent. Can we tidy this up?

subprojects/degrees-and-programs/source/cascade/formats/Degrees-programs/PrimaryContent.vm
Sean-BREI: A randomly generated number helps create a unique identifier for each of the interest's corresponding label and input field in order to avoid conflict with other interests' labels and input fields.

For example, we have two interests with a value of "Science & Technology." These values are used and mutated to help create an id for the input field: E.g. "dap-discover-interest-${rand}-science-and-technology". However, without a randomly generated number in conjunction, ids will be duplicated when values are similar and create a conflict with Accessibility.

Browsers are forgiving when rendering HTML, but duplicate ids can create potential problems with behavior as well. So, this addition will help prevent any issues with JavaScript in the future.

sharissis-brei:
@tatwell Could you clarify what you mean by the "lorem ipsum" issue?

subprojects/degrees-and-programs/source/cascade/formats/Degrees-programs/PrimaryContent.vm
tatwell: Thanks for explanation. Can you add that as a comment to code? First paragraph will be sufficient:

    // Random number helps create a unique ID for each motivation's corresponding label and input field. 
    // Same with interests below.

tatwell:
@sharissis-brei In [subprojects/degrees-and-programs/build/assemble/fixtures/dap-feature.json](https://github.com/chapmanu/cascade-assets/pull/182/files#diff-f528bd010616aad92b1901a7f054973bR236), it looks like same "lorem ipsum" value gets replaced a dozen times. At a glance, it looks like an obvious violation of the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

It's not clear to me why it needed to be changed. Is this test data or does it serve a functional role (e.g. establishing formatting then getting swapped out by a ajax call)?

sharissis-brei:
@tatwell That is all test data. This change was made by @Sean-BREI for testing purposes. This content will not output to any of the assets for cascade – it's just used for running the project locally.

tatwell:
@sharissis-brei Thanks for clarifying. Then I'm fine with that as is.

sharissis-brei:
@tatwell I just added the comments mentioned above. Let me know if you need any other changes.

sharissis-brei:
@tatwell I've added the font size increase for the results to this PR. Please let me know if there is anything else you'd like me to take care of before you merge this in.

PR #183, opened 2017-05-05
[DEFECT] On www site, search results block doesn't behave nicely.

ziegl107:
https://trello.com/c/84FugV7d/350-defect-on-www-site-search-results-block-doesn-t-behave-nicely

### DIFF
--- a/app/assets/stylesheets/omni_nav/modules/omni-nav.scss
+++ b/app/assets/stylesheets/omni_nav/modules/omni-nav.scss
@@ -175,6 +175,7 @@ html, body { margin:0; padding: 0;}
 
 
         &.expanded {
+            // value is related to cu-search-open-trigger so search icon overlaps login on mobile
             z-index: 99998;
             div.cu_dropdown_menu {
                 max-height: 200px;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/omni_nav/modules/omni-nav.scss
```diff
@@ -176,7 +175,7 @@ html, body { margin:0; padding: 0;}
 
 
         &.expanded {
-
+            z-index: 99998;
```
tatwell: Include a comment here explaining this value. As a rule, try to include comments for magic values like this. Here's a little more on magic numbers:

http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

app/assets/stylesheets/omni_nav/modules/omni-nav.scss
```diff
@@ -976,11 +1009,11 @@ a.cu_nav_button {
 }
 
 .cu-search-open-trigger {
+    z-index: 99999;
```
tatwell: I assume this value is connected with previous one. Add comment here noting that.

mandythomas:
@tatwell in the omni-nav.scss I see Kayla added some @media statements to specify a max-width size rather than using the bp() functions from base\mixins.scss (or creating a new mixin in that file then referencing it in omni-nav.scss). Is that ok or do you want that fixed for consistency?

tatwell:
@mandythomas is right. We should use SASS settings to keep things consistent. Thanks for pointing that out, Mandy.

mandythomas:
looks good

PR #184, opened 2017-05-08
Merging dev into master

waidelacruz:


mandythomas:
@waidelacruz and @tatwell this merge by Wai from Dev to Master included some things from Sean Stacho at BREI related to the campus map. Is that ok/intended?

waidelacruz:
@mandythomas @tatwell No that wasn't intended, oh my! Tom, is it safer to deploy from branch instead of issuing a PR from Dev to Master?

tatwell:
@mandythomas @waidelacruz It's ok. This is expected behavior. Sean's changes did not conflict with Wai's. I merged his PR in recently but it never got merged from development into master. And now it did.

Wai, when you do a development to master PR, I recommend looking over the PR to see if there are any unexpected changes included in it, like this. If you see something unexpected and aren't sure what to do, just let me know and I'll double check.

In this case, it's no problem.

waidelacruz:
@tatwell OK, good to know, whew! Sorry for the oversight! Glad you noticed this @mandythomas!

PR #188, opened 2017-05-12
Switched to Cascade build (was using local build)

sharissis-brei:
@tatwell I forgot to switch from the local build to the Cascade build in the last PR. Please merge this in to prevent the wrong assets from being deployed to the CMS.


tatwell:
@sharissis-brei No problem merging this. But in response to your comment:

> Please merge this in to prevent the wrong assets from being deployed to the CMS.

This will have no effect on the assets deployed by the asset build script. It only includes files in the `app/assets` directory.

sharissis-brei:
@tatwell I guess I was mistaken. I wanted to be sure nothing for the local configuration would get deployed by accident. Thanks for the clarification!

PR #196, opened 2017-06-06
[NO MERGE] As an analytics tracker for the ad landing page, I would like to track the conversions in Google Analytics. (Spike)

tatwell:
Trello Card: https://trello.com/c/EihkRp2T

This is a spike demonstrating how tracking goal can be accomplished. There will be a separate story to implement this properly. This will serve as a reference for that story.

When that story is complete, close this without merging.


kaylaziegler:
Product owners might not be moving forward with this implementation of GA tracking. Closing for now. We can still reference later if necessary.

PR #198, opened 2017-06-09
Global nav off canvas icon update

waidelacruz:
Updated icons on Global and Off-Canvas Nav.


waidelacruz:
@tatwell Added new changes to Global SubNav font stack and background color.

Issue #200, opened 2017-06-14
WAVE tool reports redundant title text in OmniNav navbar.


tatwell:
Caused by conflicting S&C and Blogs tab titles in off-canvas nav. Example:

```
<!-- under collapsible School & Colleges tab -->
<a href="business/index.aspx" title="Argyros School of Business &amp; Economics">Argyros School of Business &amp; Economics</a>

<!-- under collapsible Blogs tab -->
<a href="https://blogs.chapman.edu/business/" title="Argyros School of Business &amp; Economics">Argyros School of Business &amp; Economics</a>
```

Will need to be fixed both in Cascade and OmniNav builder.

Trello card: https://trello.com/c/xoZhXFCI

tatwell:
This will be released together with these stories once they are complete:

- https://trello.com/c/wLq8VOx0/167-as-a-mobile-user-of-www-site-i-want-a-responsive-design-for-the-2-and-3-column-pages
- https://trello.com/c/GTvE6Mdy/419-urgent-as-a-content-editor-id-like-for-the-table-borders-to-show-in-the-collapsible-widget

PR #202, opened 2017-06-16
In OmniNav off-canvas nav, fix ADA issue caused by conflicting link titles.

tatwell:
- Issue #200 
- Trello Card: https://trello.com/c/xoZhXFCI

The issue was a little different than we originally thought. It's not that we had duplicate title for different elements. According to WAVE the issue is that we have links where the title attr is the same as the label. For example:

    <a href="/" title="My Label">My Label</a>

WAVE reports screen readers don't need the title in this case and it should only be included if it offers some kind of additional advice to user.

### DIFF
--- a/lib/omninav/builder.rb
+++ b/lib/omninav/builder.rb
@@ -227,7 +227,7 @@ LOGO_HTML
                        style="outline: none;" />
               </td>
               <td class="gsc-search-button">
-                <input class="gsc-search-button" title="search" type="button" value="Search" />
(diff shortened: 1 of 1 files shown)


tatwell:
Rebuild and release will wait until next story is complete:

- https://trello.com/c/GTvE6Mdy/419-urgent-as-a-content-editor-id-like-for-the-table-borders-to-show-in-the-collapsible-widget

PR #209, opened 2017-06-28
As a WWW user, I would like to be able to add a caption to the branded masthead.

tatwell:
- Trello Card: https://trello.com/c/EL4thcb6


tatwell:
@mandythomas It's not a problem. It's not really functional although I believe I was tinkering with some classes that might make it functional in some of our samples.

Mainly, I prefer to make changes in my editor. It also gives versioning for data definitions, yes.

What I do when updating data definitions is copy the latest version from production cascade over to Cascade Assets then make my edits.

PR #226, opened 2017-07-24
For the Law footer, apply Wai's new design that is on all other pages.

kaylaziegler:
https://trello.com/c/IqodSIJC/455-for-the-law-footer-apply-wais-new-design-that-is-on-all-other-pages


app/assets/stylesheets/base/colors.scss
```diff
@@ -4,6 +4,7 @@ $red:         #98002e;
 $black:       #2f2f2f;
 $warm-gray:   #edeade;
 $cool-gray:   #ecebe5;
+$law-blue-ribbon: #002f7c;
```
tatwell: Can we just call this `$law-blue`?

app/controllers/content_types/school_home_pages_controller.rb
```diff
@@ -33,7 +33,7 @@ def slideshow
         'COMMON STYLES' => cascade_block('_cascade/blocks/html/common_styles'),
         'COMMON SCRIPTS' => cascade_block('_cascade/blocks/html/common_scripts'),
         'CUSTOM THEME' => cascade_block('_cascade/blocks/custom_themes/law'),
-        'GLOBAL FOOTER' => cascade_block('_cascade/blocks/html/law_footer'),
+        'GLOBAL FOOTER' => cascade_block('widgets/shared/law_footer'),
```
tatwell: Is the block being moved in Cascade? The idea with these controllers is that the partials hierarchy mirrors what's in Cascade.

kaylaziegler: Well in Cascade, both the global footer and the law footer are under `_cascade/blocks/html/`, but then in cascade-assets the global footer is under `widgets/shared/` and the law footer is under `_cascade/blocks/html/`

app/assets/stylesheets/base/colors.scss
kaylaziegler: Actually for this, the blue color is referenced a few times in the law theme `law.scss`, should we change all of those to use this variable?

app/controllers/content_types/school_home_pages_controller.rb
tatwell: This is ok. Don't worry about it.

app/assets/stylesheets/base/colors.scss
tatwell: That's how we should do it.

kaylaziegler:
@tatwell They do share a stylesheet, I think it's as DRY as I could make it. The only footer stylings in the law.scss theme is the ribbon color which differs from the global footer.scss. I was able to delete the whole law_footer.scss stylesheet

app/controllers/content_types/school_home_pages_controller.rb
mirabelrice: This is probably okay for now, in the future i think they should both be under `_cascade/blocks/html/` if thats how it is in cascade also because I don't think that the footer is technically a widget?

kaylaziegler: It's just that the way cascade-assets is set up right now, everything is under `widgets`. I personally think at the moment, the law header and footer don't make sense under `_cascade/blocks/html` and for me it would be an unintuitive place to look for it given where the footer and other page elements are located hierarchically in cascade-assets.

kaylaziegler:
@mandythomas So are you okay with them both being in the same subfolder under `widgets/shared` in cascade-assets? I definitely wouldn't move anything in Cascade CMS, we're just talking about where the two are located in cascade-assets

mandythomas:
@kaylaziegler yes I'm ok with them both being in same folder in cascade-assets. Makes sense so we know they're related

mandythomas:
@kaylaziegler actually I think I spoke too soon. I agree they should be together (somewhere) but why not have them both in a folder that matches Cascade CMS hierarchy closer. The footer isn't really a widget. Wouldn't they both be better moved into _cascade/blocks/html/ instead of the other way around?

mandythomas:
Sorry I keep seeing things after I clicked Approve that I want to comment on. in the _law_footer.html which is the markup plus links, you have the literal internal links from Cascade (or dev-cascade) eg. <a href="/entity/open.act?type=page&amp;id=bfb0b918c04d744c6ab45b2981d655c9" target="_parent" title="Library">Library</a>

For this static code in cascade-assets those should all be changed to a path that when this is copied to Cascade, will automatically set those entity/open.act links to the right internal link. If you leave it the way it is, you may be putting an id=xyz that worked on dev into production and there the page isn't that same id number.  The link for Library for example needs to be /law/student-resources/library/index.  You probably copied from the XML view in dev-cascade instead of going into the HTML source and taking that when you brought over the current content to cascade-assets. That way the code form cascade-assets is way more ready to copy to Cascade CMS, almost as is, except for the path to weather icon.

mandythomas:
@tatwell to answer one of your questions above about why there are still 2 blocks in Cascade (one for main and one for footer) is that these blocks are a mix of html markup and content (the links and link labels) so there has to be a Law one that has their unique links. The markup and css however for the footer is now at least all using the same classes and a shared css file since @kaylaziegler is removing the law-specific css for footers now

kaylaziegler:
@mandythomas I've changed the location of the footers to `_cascade/blocks/hmtl/` like they are in Cascade and fixed the internal links if you want to review it again

app/assets/stylesheets/base/variables.scss
```diff
@@ -111,5 +111,8 @@ $wilkinson-primary: #008ec1;
 $wilkinson-secondary: #209cc9;
 $wilkinson-tertiary: #007ba7;
 
+$law-primary: #002f7c;
+$law-secondary: #01235b;
+
```
tatwell: Any idea what `primary` and `secondary` mean in practice?

kaylaziegler: According to Wai, `primary` is the main color like for the ribbon, `secondary` is the hover color, and `tertiary` is the active color.

mandythomas: a little bit more about the primary vs secondary. Most schools just use our black ribbon so where the pri and sec colors really come into play are in the background colors of a few widgets that we change (theme) for each of the 10 schools such as the buttons and callouts in the right column. If you look at https://www.chapman.edu/wilkinson/art/index.aspx the primary color is the medium Cyan on the buttons and follow us and masthead. Non-schools would have red in those places. The secondary is then the hover on the buttons and tertiary the active state when you click it. As Kayla said, but as applies to widgets not the ribbon in the footer

PR #227, opened 2017-07-26
Adjusts width and padding for right and middle columns in 3-column te…

khafa100:
Story: https://trello.com/c/KeaedmoK

Just adjusts the width of the right column and the padding of the middle column.

### DIFF
--- a/app/assets/stylesheets/layout_old/middle-right-container.scss
+++ b/app/assets/stylesheets/layout_old/middle-right-container.scss
@@ -1,5 +1,6 @@
 .middleRightContainer {
   text-align: left;
+  padding: 0 10px;
 
   @include media($medium-screen) {
     position: relative;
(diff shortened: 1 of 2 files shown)


kaylaziegler:
@khafa100 @tatwell @mandythomas @mirabelrice Is there a specific reason we set the right column `max-width: 206px` instead of `width: 206px` in this pull request?

khafa100:
@kaylaziegler If I remember correctly, it was because leaving it as `width: 206px` would override another css property that was `width: 100%`. You don't see it in this pull request, but if you inspect the right column widget in the sample webpage in this story you'll see what I'm talking about. So I decided it was better to just set a max limit to ensure consistency in the future.

kaylaziegler:
@khafa100 `width: 206px` doesn't override the `width: 100%` property of the photo image tag.  `max-width` instead of `width` implies that it's possible for the right column to be less than 206px, but that can never be the case (when screen size > 1024px). It should always be 206px. That's the issue we're seeing in this story: https://trello.com/c/jnlQkoKK

khafa100:
@kaylaziegler Oh, I see. I guess I misunderstood. I thought the combination of the `width: 100%` and the `max-width: 206px` would work together to prevent future column issues. This one is my bad.

PR #230, opened 2017-08-11
matched the three column spacing to the two column

mirabelrice:


mandythomas:
@tatwell curious what you guys think of this... that as we modify various files that are currently in the layouts_old subfolder, we move them to another folder with a name that makes sense (because at least in this moment we've identified that it is a file in use and not truly "old"). Will make the bigger cleanup later easier if they're aren't so many files to go thru in the layouts_old folder to see which are in use and which may be obsolete. just a thought

tatwell:
@mandythomas I like that idea of getting rid of `layouts_old` directory but I think it should be done as its own story. The fear I have in trying to do it piecemeal is it leads to inconsistency or incoherence. It also opens up the possibility of complications and scope creep in small stories like this.

PR #232, opened 2017-08-23
Ford Attallah College of Educational Studies name change

mandythomas:
Changed themes, names of school icons and logos for off-canvas nav, filter logic for faculty directory. Did not yet change feeds or acronyms related to news and events widget. Those will happen with the blog is renamed


app/views/widgets/shared/_omninav.html
```diff
@@ -302,7 +302,7 @@
                           <a href="https://blogs.chapman.edu/business/" title="Argyros School of Business &amp; Economics">Argyros School of Business &amp; Economics</a>
                         </li>
                         <li>
-                          <a href="https://blogs.chapman.edu/ces/" title="College of Educational Studies">College of Educational Studies</a>
+                          <a href="https://blogs.chapman.edu/ces/" title="Attallah College of Educational Studies">College of Educational Studies</a>
```
kaylaziegler: @mandythomas Are we leaving the old name here because the name hasn't been changed for the blog?

mandythomas: No, thanks you're right, I missed that one. We're leaving the url of /ces because that's not changed for blogs yet but I only changed the title in the above link and forgot the visible label. Thanks I will fix

PR #241, opened 2017-09-12
Refactoring velocity formats for Left Nav

kaylaziegler:
Probably no merge for this

https://trello.com/c/CbcLWyvb/546-in-dev-cascade-rewrite-left-column-menu-block-using-faster-index-block


kaylaziegler:
File names match what's in dev-cascade right now. The lowercase and underscore do not match the current naming convention being used in `_cascade/formats/`.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
```diff
@@ -0,0 +1,73 @@
+## Format used by index pages where siblings don't want to be shown.
```
kaylaziegler: This format somewhat mimics what `left_nav_format` does, except one is for index pages and one is for non-index pages.

With the index pages, the velocity format renders only children. With non-index pages, the velocity format renders only siblings.

We currently have a separate velocity format for not showing siblings (for index pages). It works fine with the new block, so I just refactored the existing one quite a bit. On the other hand, the primary left nav velocity format does NOT work with the new block. But it will work for pages that aren't index pages. So that's how I justified making a new separate format for non-index pages to use, in addition to the new block.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
```diff
@@ -0,0 +1,164 @@
+## Format used by index pages with the brute-force block that loads the entire file tree.
```
kaylaziegler: This one is just an attempted refactor of the left nav velocity format currently in use, `_cascade/formats/level/LeftNav_Velocity`. Didn't make any changes to accommodate the new block we want to use because it doesn't contain some of the file tree information we need to make it function the same way it currently does.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
mandythomas: @kaylaziegler Hopefully I'm wrong about what I think you're proposing...  but if you're hoping to have 2 blocks and 2 formats and clients set one of the other each time they create a new page, or that we go through and change all the existing pages to use one or the other depending on if the page is index or non-index, then I don't think that's going to be a reasonable solution.  The LeftNavNoSibling is only on a select few pages and was done by the Web Team, not clients.  Don't model the new solution after that concept. A Config Set for a template like the 2 Column Modular can only have 1 block and 1 format for the LEFT NAV region and should apply to all pages no matter what page name is given to resulting page. Maybe we can talk more tomorrow

tatwell: Yes, this had crossed my mind, too. There's still value in what's presented here: it cleans up the code and makes future changes easier. 

Still, based on what @mandythomas points out, it sounds like we'll have to explore other options to find a viable general solution to the slow-loading problem.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
```diff
@@ -0,0 +1,125 @@
...
+                        </a>
+                    </li>
+                #end
+            #end
+        #end
+    </ul>
+#end
+
+## This macro sets up the siblings
+#macro ( showPage $siblings )
```
tatwell: Macro name and explanation above don't seem to line up.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
```diff
@@ -0,0 +1,125 @@
...
+            <a href="$parent.getChild("link").value" title="${parentTitle}">
+                $parentTitle
+            </a>
+            </li>
+        </ul>
+    </div>
+#end
+
+#if( $current )       
+    #set ( $siblings = $_XPathTool.selectNodes($currentPage, "//system-folder[system-page[@current and name = 'index']] | //system-page[@current and name != 'index'] | //system-folder[system-page[@current and name = 'index']]/following-sibling::*[name != 'index'] | //system-folder[system-page[@current and name = 'index']]/preceding-sibling::*[name != 'index'] | //system-page[@current and name != 'index']/following-sibling::*[name != 'index'] | //system-page[@current and name != 'index']/preceding-sibling::*[name != 'index']") )
```
tatwell: This line could use some explanation.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
```diff
@@ -0,0 +1,125 @@
...
+    #showPage( $siblings )
```
tatwell: What is `showPage` really doing here? Building the left-nav menu? Would `buildLeftNavMenu` be more descriptive?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
tatwell: So this would be the new default? And even if it doesn't fix the slow-load issue, you said that you still managed to cut the code in half, yes?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
```diff
@@ -0,0 +1,164 @@
...
+
+#if ( $_XPathTool.selectSingleNode( $contentRoot, "//system-folder[@current = 'true']/system-page[@current = 'true']") )
+    #set( $current = true )
```
tatwell: A little confused by `$current` here. In what case might `$current` be false?

I guess one case is the one I was dealing with yesterday when I was working with a page in the `test-section` folder that was marked hidden.

Still, it's a little bit of a struggle to make sense of this here. Anything that can make it clearer (code comment, different var name) will be appreciated by future me.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
```diff
@@ -0,0 +1,164 @@
...
+            <ul>
+                <li class="title">
+                <a href="[system-asset]${parentPath.getChild("link").value}/index[/system-asset]" target="_top" title="${parentTitle}">
+                    $parentTitle
+                </a>
+                </li>
+            </ul>
+        </div>
+    #end
+#end
```
tatwell: What does this block do?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
```diff
@@ -0,0 +1,164 @@
...
+
+#if ( $current )
+    #set ( $siblings = $_XPathTool.selectNodes($currentPage, "//system-folder[system-page[@current and name = 'index']] | //system-page[@current and name != 'index'] | //system-folder[system-page[@current and name = 'index']]/following-sibling::*[name != 'index'] | //system-folder[system-page[@current and name = 'index']]/preceding-sibling::*[name != 'index'] | //system-page[@current and name != 'index']/following-sibling::*[name != 'index'] | //system-page[@current and name != 'index']/preceding-sibling::*[name != 'index']") )
+    #showPage( $siblings )
+#end
```
tatwell: This is like the `main` routine that builds the menu, yes? Is the previous block just the header?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
```diff
@@ -0,0 +1,73 @@
...
+## This loads all children of the index page (folders and pages).
```
tatwell: If each format is intended to work with a specific block, maybe we should as a convention note that in header comment.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
```diff
@@ -0,0 +1,73 @@
...
+                    #if( $hasUnhiddenPages )
+                    </li>
+                    </ul>
+                    #end
+                </li>
+            #end
+        #end
+    #end 
+#end
+</ul>
```
tatwell: I like the idea of `main` block like above. Something like:

```
##
## Main
##
#set ( $isIndexPage = $currentType == 'index' )

# if ( $isIndexPage )
  #buildNoSiblingLeftNav()
# else
  <!-- Don't build no-sibling left nav on non-index page -->
# end
```

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
kaylaziegler: We could potentially combine the `left_nav_no_siblings_format` and the `left_nav_format` and then just have two.

I didn't know we were not okay with having more than one format the way we do now. That was never brought up. The LeftNavNoSiblings very much needed to be rewritten anyways, if we are still planning to keep it.

I can see how having more than one block might be potentially difficult to implement. It's just incredibly inefficient now and kind of ridiculous that we're using such an unnecessarily large block for all the pages that aren't indexes with siblings.

My recommendation then would be to sacrifice the functionality of those index pages that show siblings and not have the dropdowns for the siblings children, and have every left nav on every page use the smaller block I've proposed. Non-index pages and pages that use the no-siblings format would still behave the same.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
mirabelrice: A bit nit picky, but anyway to make this line more readable by using multiple lines or multiple variables?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_no_siblings_format
tatwell: On this point, I agree:

> I didn't know we were not okay with having more than one format the way we do now. 

But then discovering these sorts of things of one of the key parts of these exercises. We should try to document this somewhere where it won't be lost to future generations. :)

Which reminds me of one other suggestion I mean to make: a README file. Actually, perhaps two:

- A `README` file under `.cascade-code` that explains the purpose of this directory.
- A `README` file under `left-nav-refactor` that explain goals/motivation in general terms for this code.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
kaylaziegler: I was thinking of just changing the `showPage` macro to `displaySiblings`, because it goes through all the siblings and if the sibling has children, calls `displayChildren` macro.

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
kaylaziegler: Yes this is the new default. The names need to be altered a bit. I want to name this one left_nav_format, so I have to rename the one that's currently called that. And that one is the format to be used with the smaller block on non-index pages. It's basically siblings only, so maybe something with that? Maybe we change the other two to `left_nav_children_only` and `left_nav_siblings_only`?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
kaylaziegler: From looking at it and the output of $currentPage I'm inclined to say it would not be false ever. This format is used in a page, which means there will always be a `system-page` in the given block that has the `current='true'` attribute. Maybe @mandythomas has some insight that says otherwise?

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
kaylaziegler: Yes I'd say this was the main block. It's really just the two lines inside though now because I removed the conditional

mandythomas:
does the new format still style the leftnav menu differently if you're on an index page versus a non-index page?

kaylaziegler:
@mandythomas No, I'm just rewriting the no-siblings format and the left nav default format such that the functionality remains the same. Things sort of like renaming variables and adding comments.

I've made quite a few changes that I haven't committed yet, I'll do one soon

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_index_format
kaylaziegler: The current left nav format in production is 220 lines and the refactored one is about 130 right now

.cascade-code/left-nav-refactor/_components/left_nav/left_nav_format
kaylaziegler: This would be something Mandy might be able to explain. It's just what's used right now to get all siblings for any page no matter what its structure looks like. It's kind of complex because it has to go up one folder to get the siblings of the parent folder and all its children.

PR #242, opened 2017-09-14
As a user I want the mastheads to be ADA compliant

kaylaziegler:
https://trello.com/c/lUEU9uMD/548-as-a-user-i-want-the-mastheads-to-be-ada-compliant

### DIFF
--- a/.cascade-code/_cascade/formats/level/Masthead.vtl
+++ b/.cascade-code/_cascade/formats/level/Masthead.vtl
@@ -1,0 +1,319 @@
+#import( "/_cascade/formats/helpers.velocity" )
+
+#set ( $currentPage = $_XPathTool.selectSingleNode($contentRoot, "//system-page[@current]") )
+#set ( $masthead = $currentPage.getChild('system-data-structure').getChild('masthead') )
+#set ( $mastheadType = $masthead.getChild('mastheadType').value )
+#set ( $brandedMastheadDisplaysImage = $masthead.getChild('branded-masthead').getChild('display-image').value )
+#set ( $displaySlider = $masthead.getChild('slider').getChild('display-slider').value )
+#set ( $showMasthead = $masthead.getChild('showMasthead').value )
+#set ( $showImage = $masthead.getChild('showImage').value )
+#set ( $showBoxes = $masthead.getChild('boxes').getChild('show').value )
+
+#if ( $mastheadType == 'No Masthead' )
+   <!-- mastheadType: "No Masthead" -->
+
+## Branded Masthead design added Nov 2016.
+#elseif ( $mastheadType == 'Branded - New' )
+    <!-- mastheadType: "Branded - New" -->
+    #set ( $brandedMastheadHeader = $masthead.getChild('branded201611').getChild('header').value )
+    #set ( $brandedMastheadImagePath = $masthead.getChild('branded201611').getChild('image').getChild('path').value )
+    #set ( $brandedMastheadAltText = $masthead.getChild('branded201611').getChild('altText').value )
+    #set ( $brandedMastheadPhotoCaption = $masthead.getChild('branded201611').getChild('photoCaption').value )
+
+    <div class="masthead-branded-v201611">
+        <div class="grid">
+          <div class="column image">
+            <img alt="${_EscapeTool.xml($brandedMastheadAltText)}"
+                 src="[system-asset]${brandedMastheadImagePath}[/system-asset]"/>
+
+            #if ($brandedMastheadPhotoCaption != '')
+              <div class="photo-by">
+                ${_EscapeTool.xml($brandedMastheadPhotoCaption)}
+              </div>
+            #end
+          </div>
+          <div class="column header theme-bg-color">
+            <h2 class="header">$_EscapeTool.xml($brandedMastheadHeader)</h2>
+          </div>
+        </div>
+    </div>
+
+## Slider Masthead design added Nov 2016.
+#elseif ( $mastheadType == 'Slider - New' )
+    <!-- mastheadType: "Slider - New" -->
+    #set ( $slides = $_XPathTool.selectNodes($masthead, 'slider201611/slides/slide') )
+    #set ( $header = $_XPathTool.selectSingleNode($masthead, 'slider201611/header').value )
+    #set ( $firstSubtitle = $_XPathTool.selectSingleNode($slides[0], 'subTitle').value )
+    #set ( $subheaderClass = "#if($firstSubtitle != '')present#{else}absent#end" )
+
+  <div class="slider version-201611">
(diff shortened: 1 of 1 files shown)


kaylaziegler:
I copied this modified masthead format to Dev Cascade and I'm trying to test it out on a page but the publish queue is kind of frozen. When I inspect the draft in Cascade, it looks like it works but I want to test it in the WAVE tool once I'm able to publish it to dev-www

.cascade-code/_cascade/formats/level/Masthead.vtl
```diff
@@ -0,0 +1,322 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+
+      <div class="column header theme-bg-color">
+        <div class="aligned">
+          <h2 class="header">${_EscapeTool.xml($header).replace("\n", "<br/>")}</h2>
+          ## for ADA compliance, only show subheader if not empty
```
kaylaziegler: Only lines/block I changed: line 93-99

kaylaziegler:
Tested on this page and verified WAVE errors no longer appear: https://dev-www.chapman.edu/academics/index.aspx

.cascade-code/_cascade/formats/level/Masthead.vtl
```diff
@@ -0,0 +1,322 @@
...
+          <div class="subheader ${subheaderClass}">
```
tatwell: Shouldn't this tag be within the if statement, too?

kaylaziegler:
So after looking into this a little more, the present/absent class that gets added to the masthead header doesn't update correctly. If you add a second slide that does have a subtitle, it keeps the "absent" class while still showing the subtitle, due to this Javascript (from `cascade-assets/app/assets/javascript/cascade/level/rotator.js`):
```
var currentSubtitle = currentSlideSubtitle(slider);

$subtitle.html(currentSubtitle)

// Must hide subheader when subtitle empty in order to hide <hr/>.
if ( currentSubtitle == '' ) {
  $subheader.hide();
}
else {
  $subheader.show();
}
```
The present/absent class you see in the Velocity format only affects the first slide when you visit the page. If there is more than one slide, then the display attribute that is supposed to be set by the present/absent class gets overwritten by the javascript and doesn't make a difference anymore. That's how we're ending up with a visible subtitle for subsequent slides inside a div with class= "subheader absent" (assuming the first slide had no subtitle). There are no comments about this, but it seems like it was done on purpose, because the subtitle and subheaderClass are only set for the first slide, not dynamically with each slide:
```
#set ( $firstSubtitle = $_XPathTool.selectSingleNode($slides[0], 'subTitle').value )
#set ( $subheaderClass = "#if($firstSubtitle != '')present#{else}absent#end" )
```
Very unintuitive. Should I work around this or completely change it to be more intuitive/cleaner?

TL;DR I need to rewrite more than we thought to get subtitles to show when there are multiple slides while being ADA compliant.

mandythomas:
Can't you ignore/remove the JS and just modify the velocity format so that it tests if there's a value in the subTitle field and if not it just doesn't output the html markup for that subtitle. It's already been tested there to be empty or not, and not written out to the page ....so the JS in the live page doesn't need to again test if the subtitle is empty and apply a hidden class (does it?)

kaylaziegler:
We do need the javascript still to dynamically change the slide subtitles. I've rewritten it as best I can think of without making any massive changes.

.cascade-code/_cascade/formats/level/Masthead.vtl
```diff
@@ -0,0 +1,323 @@
...
+          <h2 class="header">${_EscapeTool.xml($header).replace("\n", "<br/>")}</h2>
+          ## for ADA compliance, only show subheader if not empty
+          <div class="subheader">
+            ## HTML for subsequent slides comes from javascript
+            ## See app/assets/javascript/cascade/level/rotator.js
+            #if ( $firstSubtitle != '' )
+              <hr />
+              <h3>${_EscapeTool.xml($firstSubtitle).replace("\n", "<br/>")}</h3>
+            #end
+          </div>
```
kaylaziegler: even this if statement can be moved to the javascript if we want. I just left it here because that's how it was before and added a comment about it.

.cascade-code/_cascade/formats/level/Masthead.vtl
kaylaziegler: @tatwell it doesn't need to be removed for ADA compliance, and the javascript wouldn't work without it

.cascade-code/_cascade/formats/level/Masthead.vtl
tatwell: The `replace` code I recognize as my own. This was to accommodate a product owner request to be able to control line breaks in subtitles. 

I probably should have added a comment for it. If we do move it to js, let's definitely add comment.

kaylaziegler: I'm indifferent when it comes to where we should put this conditional for the first slide, we can either leave it here or move it to the javascript. Idk if anyone has any strong opinions about it.

app/assets/javascripts/cascade/level/rotator.js
```diff
@@ -37,11 +37,13 @@ $(function () {
...
 
             var currentSubtitle = currentSlideSubtitle(slider);
+
+            // Always remove contents of div when switching slides
             $subheader.empty();
 
             // Remove contents of subheader div if no subtitle present
             if ( currentSubtitle != '' && currentSubtitle !== undefined ) {
-                $subheader.append("<hr /><h3>"+currentSubtitle+"</h3>");
+                $subheader.append("<hr /><h3>"+currentSubtitle.replace(/(\n)+/g, "<br/>")+"</h3>");
```
tatwell: Isn't this being done in Cascade? I'm worried we not establishing a consistent practice here: sometimes we do it in Cascade (backend), sometimes in frontend. 

Also, if we do decide to do it here, I'd prefer this was done on its own line:

    // This allows users to insert their own linebreaks in the subtitle.
    currentSubtitle = currentSubtitle.replace(/(\n)+/g, "<br/>")

(`replace` returns a string?)

kaylaziegler: No, this has to be here in the Javascript. The change in subtitle is associated with a click event, and we don't use formats for that. The only one that can be set by the format is the very first slide subtitle, when the page first loads.

tatwell: Ah, I see. Can you note this in a code comment here?

kaylaziegler: There's already a comment in the velocity format about where the subtitle gets set in the javascript but I can add one to the javascript

PR #244, opened 2017-09-18
As a web team member, I want the ShareThis widget removed from www pages.

kaylaziegler:
This is just making sure cascade-assets and Cascade are in sync for removing the Share This widget.

https://trello.com/c/VhpKeVbf


static/_files/css/level_2013.css
```diff
@@ -2529,16 +2529,6 @@ html{
...
-/*Middle Column*/
-
-.addThis{
-    clear:both;
-    margin: 20px 0 10px;
-}
-
-.addThis .twitter_share { top: 2px; position: relative; float: left; }
-.addThis .fb_like_container { top: 2px; float: left; position: relative; width: 110px !important; overflow: hidden; }
-
```
kaylaziegler: not sure if this file change is necessary, do we actually use anything in the `static` folder? I couldn't figure it out

mandythomas: I'm not sure either as we found 2 versions in Cascade as well, one original static and another within the velocity format so the Yes/No could be tested. I think it best to clean it up everywhere.

kaylaziegler:
Any other comments on this? See trello card comments for release steps in Cascade

PR #245, opened 2017-09-19
[Spike] Replace dynamic menus index block with locator tool.

tatwell:
Trello Card: https://trello.com/c/InwKZ4w9

I believe we can use the Locator tool in conjunction with the $currentPage Cascade API object to build contextual nav components more efficiently.

Demo page with sample data:

- https://dev-www.chapman.edu/test-section/tatwell/query_api_spike/demo.html.aspx

Dev Cascade page (where you can check metrics):

- [Chapman.edu/test-section/tatwell/query_api_spike/demo.html](https://dev-cascade.chapman.edu/entity/open.act?id=96aa7446c0a814531e0ec9b7fcfd2951&type=page)

Feel free to raise question or suggest other trials.

### DIFF
--- a/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
+++ b/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
@@ -85,8 +85,8 @@
 
   # ParentSiblings: title : visible?
   #foreach( $parentSibling in $grandParentFolder.children )
-    #set ( $hideSettingSelector = "dynamic-metadata[1][value='Yes'] or system-page[name = 'index']/dynamic-metadata[1][value='Yes']" )
-    #set ( $isVisible = $_XPathTool.selectNodes($parentSibling, $hideSettingSelector) == [false] )
+    #set ( $metaHide = $parentSibling.metadata.getDynamicField('Hide from navigation').value )
+    #set ( $isVisible = $metaHide != 'Yes' )
     $parentSibling.name : $isVisible
   #end
   </pre>
@@ -130,8 +130,8 @@
   <pre>
   # queryUmbrellaCategory
   umbrellaPath: $umbrellaPath
-  Sibling Count: $siblingCount of $rootFolder.children.size()
-  Child Count: $childCount of $umbrellaFolder.children.size()
+  Visible Sibling Count: $siblingCount of $rootFolder.children.size()
+  Visible Count: $childCount of $umbrellaFolder.children.size()
   Sibling Paths:
     - First: $firstSiblingPath
     - Last: $lastSiblingPath
@@ -164,7 +164,7 @@
 <div id="format-main">
 
   <h1>Query API Spike</h1>
-  <p><a href="https://github.com/chapmanu/cascade-assets/blob/spike-locator-tool/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl">query_api.vtl source</a><p>
+  <p><a href="https://github.com/chapmanu/cascade-assets/blob/spike-locator-tool/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl">query_api.vtl source</a></p>
 
   #debugCurrentPage()
   #debugPathTools()
--- a/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
+++ b/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
@@ -131,7 +131,7 @@
   # queryUmbrellaCategory
   umbrellaPath: $umbrellaPath
   Visible Sibling Count: $siblingCount of $rootFolder.children.size()
-  Visible Count: $childCount of $umbrellaFolder.children.size()
+  Visible Child Count: $childCount of $umbrellaFolder.children.size()
   Sibling Paths:
     - First: $firstSiblingPath
     - Last: $lastSiblingPath
--- a/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
+++ b/.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
@@ -22,39 +22,59 @@
 ##
 ## Macros
 ##
-#macro ( queryCurrentPageParentSiblings )
-  #set ( $parentSiblings = [] )
-
-  ## Based on https://www.hannonhill.com/kb/Script-Formats/#locator-tool
-  ## TODO: walk slashes left of current path to get grandparent
+#macro ( debugPathTools )
   #set ( $currentPath = $currentPage.getChild('path').value )
-  #set ( $grandParentFolder = $_.locateFolder('/test-section/tatwell') )
+  #set ( $pathList = $currentPath.split('/') )
+  #set ( $parentIndex = $pathList.size() - 2 )
+  #set ( $parentFolderName = $pathList.get($parentIndex) )
+  #set ( $grandParentPath = $_StringTool.substringBefore($currentPath, $parentFolderName) )
 
   <pre>
+  # debugPathTools
+  currentPage.parentFolder: $currentPage.parentFolder
+  isNull: $_PropertyTool.isNull($currentPage.parentFolder)
   currentPath: $currentPath
-  grandParentFolder.path: $grandParentFolder.path
-
-  # grandParentFolder
-  $_PropertyTool.outputProperties( $grandParentFolder )
-  </pre>
-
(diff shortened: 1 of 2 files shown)


.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+##
+
+
+##
+## Vars
+##
+## Cascade provides a $currentPage object to every format.
+
+## This can be selected when Page Block is present:
+##set ( $pageNode = $_XPathTool.selectSingleNode($contentRoot, "//system-page[@current = 'true']") )
```
tatwell: This is what we currently use in most cases to set `$currentPage`. Two drawbacks to this:

- It requires a block such as Page Block.
- It overwrites the `$currentPage` Cascade object that Cascade provides for free in every format.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+
+  #set ( $currentPath = $currentPage.path )
+  #set ( $pathList = $currentPath.split('/') )
+  #set ( $parentIndex = $pathList.size() - 2 )
+  #set ( $parentFolderName = $pathList.get($parentIndex) )
+
+  ## locateFolder fails with trailing slashes. Thus the leading slash added here.
+  ## This will also need some sanity checks for paths less than 3 folders deep.
+  #set ( $grandParentPath = $_StringTool.substringBefore($currentPath, "/${parentFolderName}") )
+  #set ( $grandParentFolder = $_.locateFolder($grandParentPath) )
```
tatwell: Notice how we get to parent siblings assets through grandparent. As code comment notes, we would need additional sanity checks to put into real world use.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+  currentPath: $currentPath
+  grandParentPath: $grandParentPath
+  grandParentFolder isNull?: $_PropertyTool.isNull($grandParentFolder)
+  parentSiblings: ${_DisplayTool.list($parentSiblings,"/")}
+  grandParentFolder.path: $grandParentFolder.path
+
+  # ParentSiblings: title : visible?
+  #foreach( $parentSibling in $grandParentFolder.children )
+    #set ( $metaHide = $parentSibling.metadata.getDynamicField('Hide from navigation').value )
+    #set ( $isVisible = $metaHide != 'Yes' )
```
tatwell: I'd like to believe there is a cleaner way to do this. But this seems to work.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+    $parentSibling.name : $isVisible
+  #end
+  </pre>
+#end
+
+#macro ( queryUmbrellaCategory )
+  ## For off-canvas nav
+  ## This would normally need to be extracted from currentPage's path.
+  #set ( $umbrellaPath = "/law" )
```
tatwell: As noted, some path-grinding code with sanity checks is going to be needed here in real life. It will also require our hardcoded `$umbrellaCategories` list.

tatwell:
Reminder to please review this PR. It's an important one as it proposes a new paradigm for Cascade development and will hopefully solve our slow page loading issue.

.gitignore
```diff
@@ -17,6 +17,7 @@
 
 # Untracked Directories
 .dev/
+.pseudo/
```
kaylaziegler: Is there a  reason for adding this?

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+#end
+
+#macro ( queryCurrentPageParentSiblings )
+  ## Note: Velocity (Java) arrays don't have a pop or remove method.
+  #set ( $parentSiblings = [] )
+
+  #set ( $currentPath = $currentPage.path )
+  #set ( $pathList = $currentPath.split('/') )
+  #set ( $parentIndex = $pathList.size() - 2 )
+  #set ( $parentFolderName = $pathList.get($parentIndex) )
```
kaylaziegler: Will this ever throw an index-out-of-bounds type of error?

.gitignore
tatwell: Yes, I had started by stubbing out a pseudocode version which I put in here. I guess I could have put it in the `.dev` folder. I was going to add a comment but I figure once you know that it is a folder for actual pseudo-code, it's self-explanatory.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
tatwell: Yes, this is good example of what the comment about needing sanity checks was referring to. I don't think it's needed for the spike but, yes, it will definitely be needed in actual implementation.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
```diff
@@ -0,0 +1,175 @@
...
+  \- isNull: $_PropertyTool.isNull($currentPage.parentFolder)
+  currentPath: $currentPath
+  pathList.size: $pathList.size()
+  parentFolderName: $parentFolderName
+  grandParentPath: $grandParentPath
+  </pre>
+#end
+
+#macro ( queryCurrentPageParentSiblings )
+  ## Note: Velocity (Java) arrays don't have a pop or remove method.
```
tatwell: This comment is specifically true of the arrays return by `split`, not array lists you may create yourself. For example, this list will support `remove`:

    #set( $myList = [1,2,3] )

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
mandythomas: I'm not sure that's quite true. For purposes of the LeftNav_velocity format we use the big LeftNav index block and from that we get the current page thru the variable you note above. We don't do a second call to another block such as Current Page in order to get that. A format can only act on one block at a time if using XPath

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
mandythomas: what if the page has no grandparents because it's in a folder 1 level off the root? Wouldn't the index page for /admission folder, for example, have the CU homepage as parent but no grandparent?

tatwell: Yes, this will need to be part of the sanity checking mentioned in the code comment.

.cascade-code/Chapman.edu/test-section/tatwell/query_api_spike/query_api.vtl
tatwell: I noticed this with some of the folders when I was testing: the folder will be visible but the index hidden.

Is there any reason we can't require the folders to be marked hidden in Cascade?

I think that convention would be simpler and more straightforward than having the code do multiple checks.

mandythomas: We do tell people to mark the folder as hidden but they may or may not always do that, opting to mark the index page instead (or as well). The current code was written to catch both cases. Marking the index page would be the same (but bad practice) as marking the folder, indicating the whole folder should still be hidden (if you can't see the landing page you shouldn't see the children). We could see if we could generate a report of folders and pages marked as hidden and do a big cleanup so it's all done correctly, and then when your program is modified it could work as you suggest. Anyone doing it wrong going forward would see that the intended landing page didn't get hidden as they thought and fix quickly (I would think). We should run this by Ross to make sure he's ok with that change in behavior (which really was the original intent but people are people and don't always do what we'd prefer

tatwell: I like this idea. With the Off Canvas Nav, it should be simpler as it only looks at Umbrella Category folders.

With the Left Nav, yes, I'd prefer to check only folders. But since this involves more end-user responsibility, I'd be more flexible with including the index file check there. We can look it at more closely when we get to that story.

PR #251, opened 2017-09-28
Added missing alts to Flickr photos

sharissis-brei:


mandythomas:
Sarah, how did this task come about? I didn't think you guys were working on any ADA things yet. Erika was going to send us a list of your new findings, and then I sent her a list of what we're working on so we don't all do the same things. We are thinking of retiring the Flickr widget but not sure. I'm surprised you are already working on ADA list.

Issue #252, opened 2017-09-28
WAVE reports Skipped Level Heading alert for News/Events widget.


tatwell:
Trello Card: https://trello.com/c/UVkFhbQx

Examples:

- https://dev-www.chapman.edu/campus-services/marketing-communication/guidelines-and-resources/web/primary-featured-news-events.aspx
- https://dev-www.chapman.edu/campus-services/marketing-communication/guidelines-and-resources/web/left-col-featured-news-events.aspx

Screenshot:

![screen shot 2017-09-28 at 2 19 11 pm](https://user-images.githubusercontent.com/12798918/30990986-0e0e1e80-a458-11e7-921c-c49d902f2bfd.png)

tatwell:
Released refactored formats with Cascade assets. Republished following pages and verified issue fixed:

- https://www.chapman.edu/campus-services/marketing-communication/guidelines-and-resources/web/primary-featured-news-events.aspx
- https://www.chapman.edu/campus-services/marketing-communication/guidelines-and-resources/web/left-col-featured-news-events.aspx

PR #253, opened 2017-09-29
As a user of the WWW site I want the News/Events widget to be accessible.

tatwell:
Trello Card: https://trello.com/c/UVkFhbQx

Issue: #252

### DIFF
--- a/.cascade-code/_cascade/formats/modular/widgets/Featured_News_Events_Left_Col
+++ b/.cascade-code/_cascade/formats/modular/widgets/Featured_News_Events_Left_Col
@@ -30,6 +30,7 @@
     <a class="no-js-link" href="https://blogs.chapman.edu/happenings/" title="View All News Stories">View All News »</a>
     <a class="no-js-link" href="https://events.chapman.edu" title="View All Event Listings">View All Events »</a>
 
+    ## Featured Tab Content
     <li class="active">
       <div class="featured">
         <img src="" alt="Chapman University" class="image" height="130"  width="225" />
@@ -46,6 +47,7 @@
       </div>
     </li>
 
+    ## News Tab Content
     <li>
       <div class="news clearfix">
         <img class="loading" src="/_files/level/img/ajax-loader.gif" alt="Loading events list..."/>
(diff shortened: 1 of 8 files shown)


.cascade-code/_cascade/formats/modular/widgets/Featured_News_Events_Left_Col
```diff
@@ -1,97 +1,89 @@
...
-                    <div class="title"></div>
-                    <div class="description">View the published page to see this section.</div>
-                    <div class="readMore"></div>
-                </div>
-            </li>
-            <li>
-                <div class="news clearfix">
-                    <img class="loading" src="/_files/level/img/ajax-loader.gif" alt="Loading news list..."/>
-                    <div class="story story1" itemscope="itemscope" itemtype="http://schema.org/Article">
+#*
```
kaylaziegler: Shouldn't these two velocity formats have the .vtl extension?

mandythomas:
@tatwell I didn't realize, or maybe I knew at one time and have forgotten, that the velocity formats are in github. They can't run in the local environment and there is version control for these in Cascade so why do we keep them in github as well making this more work to update them multiple places? I don't think I have been updating github with format changes I made in recent months. I also discovered the data definitions are in there. I haven't been updating them there either. Same question.

.cascade-code/_cascade/formats/modular/widgets/Featured_News_Events_Left_Col
tatwell: New ones should. These are updates to existing one.

tatwell:
@mandythomas The dev team just started putting formats in Github. It's the same practice as with data definitions -- except with these there is no attempt to make them functional within the Cascade Assets demos (why they're in a `.` directory).

There's a few advantages with tracking them in Github:

- Easier to share and review code (main advantage)
- With code reviews, easier to establish and promote best practices
- More full-featured version control (e.g. branching and tagging)

As with data definitions, the version in production should always be considered the master version. When devs work on a format in Github, they should always be copying the version from Cascade production as the first step. Working through this Github repo is not required, but I think it's a good idea.

Issue #239, opened 2017-08-31
As a Cascade User, I would like to see faster internal page load times.


kaylaziegler:
https://trello.com/c/1Hytt7sS/496-as-a-cascade-user-i-would-like-to-see-faster-internal-page-load-times

kaylaziegler:
Helpful Velocity Programming Docs: http://velocity.apache.org/engine/1.7/user-guide.html

Location of test blocks and formats in Dev-Cascade: `_components/system_index_test`. It contains one test block for each option that Cascade provides. The current Omni-Nav format in use right now has been copied there. There are also 2 very simple new formats I wrote that are tailored to certain blocks. They do essentially the same thing but the parsing of the blocks is a little different. These formats are in use on the following pages:

- [Test Brute-Force](https://dev-cascade.chapman.edu/entity/open.act?id=3eec04d4c0a814531b32a4ac2e76ff75&type=page)
- [Test Siblings & Render Forward](https://dev-cascade.chapman.edu/entity/open.act?id=3e4d83edc0a814531b32a4aca832439e&type=page)

Test page rendering metrics:

- Block with everything: approx. 9917 ms
- Block with only siblings and rendering forward: approx. 129 ms

kaylaziegler:
Trello stories have been spun off to rewrite formats in Cascade to work with different block:

- [In Dev Cascade, rewrite OmniNav Off-Canvas Nav using faster index block.](https://trello.com/c/v5qK35x8/542-in-cascade-rewrite-omninav-off-canvas-nav-using-faster-index-block)
- [In Dev Cascade, rewrite Global Nav using faster index block.](https://trello.com/c/J0OCb50y/545-in-dev-cascade-rewrite-left-column-menu-using-faster-index-block)
- [In Dev Cascade, rewrite Left Column Menu Block using faster index block.](https://trello.com/c/CbcLWyvb/546-in-dev-cascade-rewrite-global-nav-using-faster-index-block)

kaylaziegler:
Trying to make sense of the left nav. Thoughts:

Question: What is considered a "sibling" in relation to the page you are currently on? We've got conflicting answers to this.

My answer: If you're on the index page (index.aspx) within a folder, I basically consider it the same thing as the folder even though it's on the same level hierarchically as other pages and folders. Those other pages and folders that are next to the index are considered children of the index.

Example:
- `copa/music/index.aspx` is the parent of `copa/music/admission/` folder and `copa/music/facilities.aspx`

[...]

PR #255, opened 2017-09-29
Refactor the Left Nav menu to use Cascade current page object instead of index block.

kaylaziegler:
Story: https://trello.com/c/jRQeKw1B

Part of the fix for this issue: #239


kaylaziegler:
The only new files that need to be reviewed:

- `.cascade-code/Chapman.edu/_components/left_nav/left_nav_siblings.vtl`
- `.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl`

This is where they're located in Dev-Cascade, you can test it on pages by swapping out the formats and selecting "no block".

Reminder of our Left Nav definitions: https://github.com/chapmanu/cascade-assets/issues/239#issuecomment-328168428

.cascade-code/Chapman.edu/_components/left_nav/left_nav_siblings.vtl
```diff
@@ -0,0 +1,144 @@
...
+                <li>
+                    <a href="${child.link}/index" title="${childTitle}" class="c-f">
+                        <span class="bullet">» </span>
+                        $childTitle
+                    </a>
+                </li>
+            #end
+        #end
+    </ul>
+#end
```
tatwell: I'll do a full review when I return. In the meantime, I think the code can be reorganized to better separate some of the logic from the output. I'll use this macro as an example.

First, I would organize the macro around the child list item and call it like so: 

```
<ul class="leftNav-sub">
  #foreach ( $child in $children )
    #buildLeftNavChild( $child )
  #end
</ul>
```

Then I find it easier to read when the `buildLeftNavChild` macro is focused on a single "output" template:

```
#macro ( buildLeftNavChild $child )
  ## Data
  #set ( $childTitle = ${_EscapeTool.xml($child.metadata.displayName)} )
  #set ( $childIsVisible = true )

  #if ( $child.assetType == 'page' && ($child.name != 'index') )
    #set ( $linkClass = "c-p" )
    #set ( $linkHref = $child.link )
  #elseif ( $child.assetType == 'folder' )
    #set ( $linkClass = "c-f" )
    #set ( $linkHref = "${child.link}/index" )
  #else
    #set ( $childIsVisible = false )
  #end

  ## Output
  #if ( $childIsVisible )
  <li>
    <a class="${linkClass}" href="${linkHref}" title="${childTitle}">
      <span class="bullet">» </span>
      $childTitle
    </a>
  </li>
  #end
#end
```

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
+## Formerly LeftNavNoSiblings. Used for index pages.
+## New left nav format that works without a block
+
+#set ( $pagePath = $currentPage.path )
+#set ( $parentDir = $currentPage.parentFolder )
+
+## Builds the left navigation
+#macro ( buildLeftNav )
+    #buildLeftNavTitle
```
tatwell: I guess the parentheses aren't needed but I'd prefer macros were called with them. It's a little visual signal that improves readability:

    #buildLeftNavTitle()

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+    <ul>
```
tatwell: I'd prefer this block below were its own macro: `buildLefNavMenu` or `buildLeftNavMenu`.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+    #set ( $children = $parentDir.children )
+    #foreach ( $child in $children )
+        #set ( $isVisible = $child.metadata.getDynamicField('Hide from navigation').value != "Yes" )
+
+        ## Child is a page
+        #if ( $child.assetType == 'page' && $isVisible && ($child.name != 'index') )
+            #buildChildPage($child)
```
tatwell: This isn't really building a page but rather a link (technically a list item).

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+#set ( $pagePath = $currentPage.path )
+#set ( $parentDir = $currentPage.parentFolder )
+
+## Builds the left navigation
+#macro ( buildLeftNav )
+    #buildLeftNavTitle
+    <ul>
+    #set ( $children = $parentDir.children )
+    #foreach ( $child in $children )
+        #set ( $isVisible = $child.metadata.getDynamicField('Hide from navigation').value != "Yes" )
```
tatwell: Ideally, everything in the loop could be encapsulated in a `buildLeftNavChild` macro.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+            ## Titles for folders are set to the name of its index page
+            #set ( $childTitle = ${_EscapeTool.xml($folderIndexPage.label)} )
+            
+            ## Only if the child folder's index page is also visible
+            #if ( $visibleIndex )
+                #set ( $visibleChildren = [] )
+                #getVisibleChildren($child.children)
+                
+                ## Needs to have a dropdown if it has visible children
+                #if ( $visibleChildren.size() > 0 )
```
tatwell: Three levels of nested `ifs` is a bad sign to me.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+            ## For folders, also need to check visibility of index page
+            #set ( $folderIndexPage = $_.locatePage("${child.path}/index") )
+            #set ( $visibleIndex = $folderIndexPage.metadata.getDynamicField('Hide from navigation').value != "Yes" )
+            ## Titles for folders are set to the name of its index page
+            #set ( $childTitle = ${_EscapeTool.xml($folderIndexPage.label)} )
+            
+            ## Only if the child folder's index page is also visible
+            #if ( $visibleIndex )
+                #set ( $visibleChildren = [] )
+                #getVisibleChildren($child.children)
```
tatwell: I see the inspiration behind the pattern here (a lack of proper functions in Velocity) but I don't like it. It obscures the fact that the var `$visibleChildren` is being used as a global.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+                            $childTitle
+                            <span class="plus">+</span>
+                        </a>
+                        #buildGrandchildren($visibleChildren)
+                    </li>
+                    
+                ## If it has no visible children it's basically just a page
+                #else
+                    <li>
+                        <a href="${folderIndexPage.link}" class="s-f-idx">
```
tatwell: I assume this class name is pre-existing.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
```diff
@@ -0,0 +1,137 @@
...
+                #getVisibleChildren($child.children)
+                
+                ## Needs to have a dropdown if it has visible children
+                #if ( $visibleChildren.size() > 0 )
+                    <li>
+                        <a href="${folderIndexPage.link}" class="s-f-idx">
+                            $childTitle
+                            <span class="plus">+</span>
+                        </a>
+                        #buildGrandchildren($visibleChildren)
```
tatwell: Do we have to worry about namespace collisions when using generic names like this? I'm not sure how scoping works exactly. 

I'd prefer we used more descriptive names as a rule.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_siblings.vtl
```diff
@@ -0,0 +1,147 @@
...
+            #set ( $discard = $visibleChildren.add($asset) )
+        #end
+    #end
+#end
+
+##
+## MAIN
+##
+#setParentFolder
+#buildLeftNav
```
tatwell: To echo the sort of MVC practice used by rail, the general pattern I'd recommend here would be:

    setVars()        ## Model in MVC
    buildLeftNav()   ## View in MVC

Secondary macros can be called within these.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
kaylaziegler: I can remove the macro and just put all the code back in here.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
kaylaziegler: Do you mean `visibleChildren`? Or `buildGrandchildren`?

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
kaylaziegler: Yes the class names are all pre-existing

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
tatwell: Before doing that, I'd take a look at the bigger picture. If you restructured this macro like so:

```
#macro ( buildLeftNav )
    #buildLeftNavTitle
    <ul>
    #set ( $children = $parentDir.children )
    #foreach ( $child in $children )
      buildLeftNav($child)
```

Maybe that would help reorganize things. The logic could be contained in that macro (and that macro itself might take advantage of other macros if it makes sense and isn't getting too convoluted.)

It's true you're still going to have that block of logic somewhere. But this might help better isolate and encapsulate the logic so that it doesn't interfere with a broader view of what's happening.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_children.vtl
tatwell: Both potentially. With var names, I don't mind so much because I think it's reasonable to assume those are sanely scoped.

With macros, I worry a little more about namespace collisions. I'm not sure if this is a real concern. But so long as they don't get too long and crazy, I prefer more descriptive.

kaylaziegler:
@mandythomas This is the pull request for the left nav refactor that I was talking to you about. This is what is in Dev-Cascade right now so you can also test it out there (`Chapman.edu/_components/left_nav/`)

.cascade-code/Chapman.edu/_components/left_nav/left_nav_no_siblings.vtl
```diff
@@ -0,0 +1,171 @@
...
+## New left nav format that works without a block
+
+#set ( $parentDir = $currentPage.parentFolder )
+
+## Builds the left navigation for non-index pages (only shows the title)
+#macro ( buildNonIndexLeftNav )
+    #buildNonIndexLeftNavTitle()
+    ## Needs to have empty <ul> tags because otherwise title does not have a bottom margin
+    ## Without it, other left col widgets will show up directly under the left nav with no space
+    <ul></ul>
```
kaylaziegler: does this make sense? it's kind of an ugly weird thing that has to be here

.cascade-code/Chapman.edu/_components/left_nav/left_nav_no_siblings.vtl
```diff
@@ -0,0 +1,171 @@
...
+                ${_EscapeTool.xml($currentPage.label)}
+            </a>
+            </li>
+        </ul>
+    </div>
+#end
+
+## Builds header for non-index pages that use this format
+## Non-index pages should display the title as the page display name and nothing else
+#macro ( buildNonIndexLeftNavTitle )
```
kaylaziegler: I made two separate macros for the left nav titles, one for index pages and one for non-indexes pages, but if we think it's not really necessary then I can just have one macro that has a conditional in it that checks which kind of page it is

.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
```diff
@@ -0,0 +1,153 @@
...
+            #set ( $indexPage = $_.locatePage("${sibling.path}/index") )
+            #set ( $visibleIndex = $indexPage.metadata.getDynamicField('Hide from navigation').value != "Yes" )
+            
+            ## Only if the index page is also visible
+            #if ( $visibleIndex )
+                #buildSiblingFolderListItem($sibling $indexPage)
+            #end
+        #end
+    #end
+#end
```
tatwell: The foreach feels out of place here. The following parses a little more easily for me:

```
## Builds left navigation
#macro ( buildLeftNav )
    ## Data
    #set ( $siblings = $parentDir.children )

    ## Output
    #buildLeftNavTitle()
    
    <ul>
    #foreach ( $sibling in $siblings )
      #buildLeftNavItem($sibling)
    #end
    </ul>
#end
```

Then basically rename this macro and remove of the foreach here.

Alternately, keep this macro, but move the `ul` into it:

```
## Builds left navigation
#macro ( buildLeftNav )
    ## Data
    #set ( $siblings = $parentDir.children )

    ## Output
    #buildLeftNavTitle()
    #buildLeftNavMenu($siblings)
#end
```

I think I like that best. (Hopefully, this isn't what you originally had.)

.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
```diff
@@ -0,0 +1,153 @@
...
+    #set ( $siblingPageTitle = ${_EscapeTool.xml($siblingPage.label)} )
+
+    #if ( $siblingPage.path == $currentPage.path )
+        <li class="active">
+    #else
+        <li>
+    #end
+        <a href="${siblingPage.link}" title="$siblingPageTitle" class="s-p">$siblingPageTitle</a>
+    </li>
+#end
```
tatwell: I'm going to nitpick here again to press the idea again of isolating data from output/templating:

```
#macro ( buildSiblingPageListItem $siblingPage )
    ## Data
    #set ( $siblingPageTitle = ${_EscapeTool.xml($siblingPage.label)} )
    #if ( $siblingPage.path == $currentPage.path )
      #set( $liClass = "active" )
    #else
      #set( $liClass = "unstyled-inactive" )
    #end

    ## Output
    <li class="${liClass}">
        <a href="${siblingPage.link}" title="$siblingPageTitle" class="s-p">$siblingPageTitle</a>
    </li>
#end
```

kaylaziegler: Can we leave the class name blank for list items that aren't active or do we have to give it some value like `unstyled-inactive`?

tatwell: It triggers my OCD but we can. 😄 

...I think. The question is: what does the WAVE tool say?

kaylaziegler: WAVE doesn't get mad about it, but I'm fine with adding it

tatwell: If WAVE's ok with it, I am, too. Thanks for checking.

.cascade-code/Chapman.edu/_components/left_nav/left_nav_no_siblings.vtl
tatwell: Ok with me. Especially with commment.

mandythomas: I checked if empty UL tags cause any accessibility issues and they don't

.cascade-code/Chapman.edu/_components/left_nav/left_nav_no_siblings.vtl
mandythomas: I like keeping 2 macros. Even though some repeat code in them it's probably easier to get your brain around them and easier to see differences if they're kept separate.  Another thing in this same section of the code.. You're putting the same verbiage in the title parameter of the menu links as is in the visible link text. I think that is actually worse for accessibility than better. The title is supposed to supplement or give you something different, or if multiple links on page share the same link text but go to different urls then use title to differentiate. I think we probably need to remove OR you can concatenate the users's label with static text such as Click on menu to go to page about ....  I can see this helping if an item is in left nav and a breadcrumb link. I read about on https://www.w3.org/TR/WCAG10-HTML-TECHS/#link-text

mandythomas: I could be wrong but I *think* that some of the classes in here eg, s-f-idx, c-f and c-p are not really for styling but codes I'd added once so I could look at the html source and see what a link in the menu represented eg I think c-f was child and a folder (index) and c-p was a chid and a page. We can keep or leave but I don't think you'll find any css styles related to them. Was when I was trying to rewrite this leftnav (several times) in the past and trying to sort out which sections of velocity produced which parts of final menu.

tatwell: If these classes are non-functional then I agree: remove them.

tatwell: > You're putting the same verbiage in the title parameter of the menu links as is in the visible link text. I think that is actually worse for accessibility than better. The title is supposed to supplement or give you something different, or if multiple links on page share the same link text but go to different urls then use title to differentiate. I think we probably need to remove OR you can concatenate the users's label with static text such as Click on menu to go to page about...

I talked with Kayla about this in standup this morning. I agree that we should probably fix this. But this is existing behavior so changing it now is beyond the scope of this story. It will have to be written up as a new user story.

mandythomas:
@Tom as far as writing up new story to change the title parameter on the links. It's your call BUT I'd suggest that because these formats are a complete rewrite anyway then it's reasonable to modify that especially when it's an accessibility issue that we can knock off now vs later, and because it's such a small item.

tatwell:
@mandythomas You can add the accessibility issue as an acceptance criteria to this release story:

- https://trello.com/c/4ekeO2Ms/600-release-leftnav-refactor-to-production-cascade

Issue #254, opened 2017-09-29
WAVE reports empty link for social icons in featured stories widget footers.


tatwell:
Trello Story: https://trello.com/c/JlStwNlL

Example:

https://www.chapman.edu/

Screenshot:

![screen shot 2017-09-29 at 1 13 13 pm](https://user-images.githubusercontent.com/12798918/31034233-0b52f34e-a518-11e7-8c30-a1fe82621a63.png)

tatwell:
Applied this fix:

- https://stackoverflow.com/a/39919878/6763239

Production release complete and fix verified with WAVE tool:

- https://www.chapman.edu/index.aspx

PR #256, opened 2017-10-03
Issue 254: Empty Links in Hompage Blogs Stories widget.

tatwell:
Trello Story: https://trello.com/c/JlStwNlL

Issue #254.

### DIFF
--- a/.cascade-code/_config/blogs.vm
+++ b/.cascade-code/_config/blogs.vm
@@ -15,10 +15,10 @@
 #set ( $BLOGS_CONFIG = {
   'academics' : [ 'Higher Ed and Technology: Academics at Chapman', '' ],
   'alumni' : [ 'Chapman Alumni', '' ],
-  'business' : [ 'business', 'Argyros School of Business and Economics', '' ],
-  'education' : [ 'education', 'Attallah College of Educational Studies', '' ],
+  'business' : [ 'Argyros School of Business and Economics', 'business' ],
+  'education' : [ 'Attallah College of Educational Studies', 'education' ],
   'collections' : [ 'Art Collections', '' ],
-  'communication' : [ 'communication', 'School of Communication', '' ],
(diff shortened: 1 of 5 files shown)


mandythomas:
the url for the blog link for Attallah College is the old one. It's now /education not /ces.

.cascade-code/_config/blogs.vm
```diff
@@ -0,0 +1,38 @@
+#*
```
tatwell: Notice these new `_config` files.

.cascade-code/_config/blogs.vm
```diff
@@ -0,0 +1,38 @@
...
+ * #set ( $blogTitle = $blogParams[0] )
+ * #set ( $blogCss = $blogParams[1] )
+ *#
+
+## blog id => [ blog title, css theme class ]
+#set ( $BLOGS_CONFIG = {
+  'academics' : [ 'Higher Ed and Technology: Academics at Chapman', '' ],
+  'alumni' : [ 'Chapman Alumni', '' ],
+  'business' : [ 'business', 'Argyros School of Business and Economics', '' ],
+  'education' : [ 'education', 'Attallah College of Educational Studies', '' ],
```
kaylaziegler: Why do some of these keys have 3 values in the array while others have 2? Also, the order seems wrong for some, like `education` and `business` should be swapped with the second element? Based on the comment at the top of the file: 
```
#set ( $blogTitle = $blogParams[0] )
#set ( $blogCss = $blogParams[1] )
```

tatwell: Yes, this was a mistake leading to some of the buggy behavior Mandy noticed. Good eye.

tatwell:
That should be the last of the changes for this story.

Call for final review. I'm planning to release tomorrow morning.

PR #260, opened 2017-10-09
Add Careers umbrella category to new OmniNav.

kaylaziegler:
Story: https://trello.com/c/SvloTLKb

The modified format can also be found/tested in Dev Cascade in `_components/omni_nav_v2/omni_nav.vtl`

### DIFF
--- a/.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav.vtl
+++ b/.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav.vtl
@@ -44,7 +44,8 @@
   [ 'research', 'Research', 'default', '/' ],
   [ 'students', 'Students', 'default', '/' ],
   [ 'support-chapman', 'Support Chapman', 'default', '/' ],
-  [ 'communication', 'SCHOOL OF COMM', 'soc', '/communication/' ]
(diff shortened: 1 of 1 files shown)


kaylaziegler:
@mandythomas No I think you might've been looking at the wrong spot, the last value in the inner arrays is always a directory, so either a specific one like `/communication/` or the default root directory `/`, which is the one I used here. The `default` value is for the class names, for communication it says `soc` not `/communication/` in the third spot.

mandythomas:
maybe. I'm kind of going fast between conference sessions but here's what I saw and at first glance that ending "/" in the 4th param would indicate to me to be a url of the root of the website not the careers section. I would have thought that would be /campus-services/career-development/  but I viewed the dev site - homepage of Careers and the menus are working, so must be fine  

[ 'communication', 'SCHOOL OF COMM', 'soc', '/communication/' ],
[ 'campus-services/career-development', 'Career', 'default', '/' ]

kaylaziegler:
@mandythomas Yes that's what it's supposed to be, as far as my understanding goes. That's the parameter that specifies where the user goes when they click on the Chapman logo in the off canvas, and it looks like for everything besides the schools/colleges we want it to go to the homepage.

PR #264, opened 2017-10-16
New omninav search

mirabelrice:


app/views/omni_nav_v2/_utility_nav.html.erb
```diff
@@ -56,11 +56,25 @@
...
+              <li class="search-filter-option"><a>Main Website</a></li>
+            </ul>
+          </div>
+        </li>
         <li class="utility-cell">
-          <%= render 'omni_nav_v2/search_box' %>
+          <div class="cu-search-box"></div>
+          <div class="search-results-container">
+            <div class="cu-search-results"></div>
+          </div>
```
tatwell: Are there any format changes associated with these changes?

app/assets/javascripts/omni-nav-v2.js
```diff
@@ -73,6 +71,211 @@ var OmniNav2 = (function() {
...
+
+      return {
+        init: initialize,
+      };
+    };
+
+    return {
+      init: initialize
+    };
+  })();
```
tatwell: Nicely organized. Much more readable the previous version. 👍

app/views/omni_nav_v2/_utility_nav.html.erb
mirabelrice: Yes there will need to be format changes for both search boxes. I will commit the updated formats before this is merged.

.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav.vtl
```diff
@@ -318,7 +319,9 @@
 
 #macro ( buildPrimaryNavSearchBox )
   <div class="nav-container primary-nav-search">
-    #buildSearchBox()
+    <div class="cu-two-column-gcs" id ="primary-nav-search">
```
tatwell: You've got an extra space in `id ="primary-nav-search"`.

More importantly, does this need an `id`? Having a child block with `id` that is same as parent's `class` name doesn't smell right.

.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav.vtl
```diff
@@ -164,6 +164,7 @@
   #buildUtilityNav()
   #buildPrimaryNav()
   #buildOffCanvasNav()
+  <div id="search-results-overlay"></div>
```
tatwell: Can you add comment explaining briefly why this is needed.

.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav.vtl
mirabelrice: Each search element needs to have a unique identifier for GCS to differentiate them so I thought it made sense to use an `id`. I should be able to refactor this to only have `id="primary-nav-search"` and remove the redundant `class`.

Issue #265, opened 2017-10-18
[URGENT] Left Nav Velocity format shows _files folders when it shouldn't and root folders show the wrong title.


kaylaziegler:
Trello Card: https://trello.com/c/zwuWdHeF

**Issue 1**: Most folders in `Chapman.edu` contain a folder called `_files` where images and other assets are stored for that folder. They should never show in the left nav, even if not marked as hidden.

**Issue 2**: Folders under the root that use the `left_nav` format are showing `${_EscapeTool.xml($folderIndexPage.label)}` as the left nav title. Ex. `Chapman.edu/360/index`.

kaylaziegler:
We have 2 left nav velocity formats: `left_nav` and `left_nav_no_siblings`.

Issue 2 isn't a problem for the `left_nav_no_siblings` format because it doesn't use any parent information to set the left nav titles, by design. It just uses the current page info.

kaylaziegler:
Pull request to fix these two issues: #266

kaylaziegler:
As a part of the fix for this issue we also added conditionals to prevent some unexpected left nav behavior caused by folders not having index pages.

Issue #262, opened 2017-10-12
[DEFECT] New OmniNav OffCanvas menu does not include all expected links.


kaylaziegler:
Trello story: https://trello.com/c/NWW6Ttr8

Some folders/pages are not showing up in the new omninav v2 off canvas menu, when they are expected to.

kaylaziegler:
The cause of this was a combination of things, but it boils down to a weird Cascade/Velocity thing. These are the offending lines:
```
#set ( $metaHideSetting = $asset.metadata.getDynamicField('Hide from navigation').value )
#set ( $isVisible = $metaHideSetting != 'Yes' )
```

The problem is with the $metaHideSetting. For some reason, the value of `$asset.metadata.getDynamicField('Hide from navigation').value` was returning "" for some folders, and returning "$asset.metadata.getDynamicField('Hide from navigation').value" for others. For the latter, Velocity prints/returns the whole function call as a string literal like that when it can't find whatever you're trying to get, so in this case it couldn't find a metadata field called "Hide from navigation". So I went to one of the folders that was returning "$asset.metadata.getDynamicField('Hide from navigation').value", I checked the box "Hide from Navigation", submitted and published it, then went back and unchecked the box and submitted/published it. Then when I checked the value of $metaHideSetting again, it returned "" instead of "$asset.metadata.getDynamicField('Hide from navigation').value". Since I had first set it to "Yes", it now had a dynamic field called "Hide from navigation". Even though I unchecked it, velocity still found the value of the field, which would be empty. At least that's how I'm explaining it to myself.

Lots of ways to deal with it, I'll comment with a solution tomorrow.

mandythomas:
that "Hide from navigation" field is a checkbox so it could be that when it's not checked the field doesn't exist, rather than existing with a value of blank. Works that way on forms on the web so maybe this is the same

kaylaziegler:
Part of the problem was also having it in two lines, as shown in the previous comment. If "$asset.metadata.getDynamicField('Hide from navigation').value" didn't return any value, it did a weird thing where it would keep the value of $metaHideSetting from the previous iteration in the loop (the previous child asset). However, this one-liner that accomplishes the same thing does work:
`#set ( $isVisible = $asset.metadata.getDynamicField('Hide from navigation').value != "Yes" )`

**Solutions**
Adding this line inside the loop before anything else would be one way to solve the unexpected behavior we're seeing: `#set ( $metaHideSetting = '' )`

[...]

PR #266, opened 2017-10-18
[URGENT] Left navigation format fixes - "_files" exclusion and root level fixes

kaylaziegler:
Trello story: https://trello.com/c/zwuWdHeF

Issue #265 

**Dev Cascade**
- Test formats: `_components/left_nav/`
- Live formats: `_cascade/formats/level/`

### DIFF
--- a/.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
+++ b/.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
@@ -25,14 +25,25 @@
 ## Builds the header for the left nav
 #macro ( buildLeftNavTitle )
     ## Data
-    #set ( $folderIndexPage = $_.locatePage("${parentDir.path}/index" ) )
+    #if ( $parentDir.path == "/" )
+        ## For folders under the root, the index doesn't need another forward slash
+        #set ( $folderIndexPage = $_.locatePage("${parentDir.path}index" ) )
+        #set ( $titleLink = $folderIndexPage.link )
+        ## This $title variable should be $folderIndexPage.label, but for the home page
+        ## that label is way too long.
+        #set ( $leftNavTitle = "Chapman University" )
+    #else
+        #set ( $folderIndexPage = $_.locatePage("${parentDir.path}/index" ) )
+        #set ( $titleLink = $folderIndexPage.link )
+        #set ( $leftNavTitle = ${_EscapeTool.xml($folderIndexPage.label)} )
+    #end
     
     ## Output
     <div class="leftTitle">
         <ul>
             <li class="title">
-                <a href="${folderIndexPage.link}" target="_top" title="${_EscapeTool.xml($folderIndexPage.label)}" >
-                    ${_EscapeTool.xml($folderIndexPage.label)}
+                <a href="${titleLink}" target="_top" title="${leftNavTitle}" >
+                    ${leftNavTitle}
                 </a>
             </li>
         </ul>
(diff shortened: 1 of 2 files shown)


kaylaziegler:
See issue #265 for the explanation about why the `left_nav_no_siblings` format didn't have the problem with the title when the format was used directly under root.

Only thing I had to change in that one was the other part of this, which was to exclude folders called `_files`.

.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
```diff
@@ -25,14 +25,25 @@
...
 #macro ( buildLeftNavTitle )
     ## Data
-    #set ( $folderIndexPage = $_.locatePage("${parentDir.path}/index" ) )
+    #if ( $parentDir.path == "/" )
+        ## For folders under the root, the index doesn't need another forward slash
+        #set ( $folderIndexPage = $_.locatePage("${parentDir.path}index" ) )
+        #set ( $titleLink = $folderIndexPage.link )
+        ## This $title variable should be $folderIndexPage.label, but for the home page
+        ## that label is way too long.
+        #set ( $leftNavTitle = "Chapman University" )
```
kaylaziegler: Discussed this with Ross. The value here is technically supposed to be "Chapman University | A Top Private University in California" because that's the display name for the home page.

That's just a bit too wordy, so Ross said "Chapman University" for the title if the parent of the current page is root.

.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
```diff
@@ -50,12 +61,13 @@
...
         ## Sibling is a folder
-        #elseif ( $sibling.assetType == 'folder' && $isVisible )
+        #elseif ( $sibling.assetType == 'folder' && $isVisible && $sibling.name != '_files' )
             #set ( $indexPage = $_.locatePage("${sibling.path}/index") )
             #set ( $visibleIndex = $indexPage.metadata.getDynamicField('Hide from navigation').value != "Yes" )
             
             ## Only if the index page is also visible
-            #if ( $visibleIndex )
+            ## Also make sure the correct index page was found
+            #if ( $visibleIndex && ( $indexPage.path == "${sibling.path}/index" ) )
```
kaylaziegler: These last two commits are to deal with folders that don't have index pages for whatever reason. They cause problems similar to what we saw in issue #262 

This is the most effective way I think to do it since velocity doesn't return anything helpful like NULL when it doesn't locate something, it leaves it as the previous value of that variable.

.cascade-code/Chapman.edu/_components/left_nav/left_nav.vtl
```diff
@@ -102,12 +103,13 @@
...
         #if ( $asset.assetType == 'folder' && $asset.name != '_files' )
-            ## Both the folder and index page must be visible
             #set ( $folderVisible = $asset.metadata.getDynamicField('Hide from navigation').value != "Yes" )
             #set ( $folderIndexPage = $_.locatePage("${asset.path}/index" ) )
             #set ( $folderIndexVisible = $folderIndexPage.metadata.getDynamicField('Hide from navigation').value != "Yes" )
             
-            #if ( $folderIndexVisible && $folderVisible )
+            ## Both the folder and index page must be visible
+            ## Also make sure this asset's index page was found correctly
+            #if ( $folderIndexVisible && $folderVisible && ( $folderIndexPage.path == "${asset.path}/index" ) )
```
tatwell: How about this approach to explain what `$folderIndexPage.path == "${asset.path}/index"` means.

```
              #set ( $folderVisible = $asset.metadata.getDynamicField('Hide from navigation').value != "Yes" )
              #set ( $folderIndexPage = $_.locatePage("${asset.path}/index") )
              #set ( $folderIndexPageExists = $folderIndexPage.path == "${asset.path}/index" )
              #set ( $folderIndexVisible = $folderIndexPageExists && $folderIndexPage.metadata.getDynamicField('Hide from navigation').value != "Yes" )

              #if ( $folderVisible && $folderIndexVisible )
```

You'd have to change it a couple other places, too.

PR #270, opened 2017-10-27
Some 3-column template pages on www are appearing with margins and column widths not consistent with the standard.

kaylaziegler:
https://trello.com/c/jnlQkoKK

See [trello comment](https://trello.com/c/jnlQkoKK/583-some-3-column-template-pages-on-www-are-appearing-with-margins-and-column-widths-not-consistent-with-the-standard#comment-59f367015d6c17c5945ab926) for explanation of issue

### DIFF
--- a/app/assets/stylesheets/layout_old/right-column.scss
+++ b/app/assets/stylesheets/layout_old/right-column.scss
@@ -78,7 +78,7 @@
     position: relative;
     display: inline-block;
     float: right;
-    max-width: 206px;
+    width: 206px;
     margin-top: 45px;
     padding-right: 0;
   }


app/assets/stylesheets/layout_old/right-column.scss
```diff
@@ -78,7 +78,7 @@
     position: relative;
     display: inline-block;
     float: right;
-    max-width: 206px;
+    width: 206px;
```
kaylaziegler: If you look at the sample pages from the story: 

- https://www.chapman.edu/students/academic-resources/tutoring-center/tutoring/index.aspx
- https://www.chapman.edu/crean/academic-programs/graduate-programs/ms-athletic-training/index.aspx

You can see the content of the right column text is pretty small. Since this setting was max-width instead of width, and the content itself didn't take up 206px, it would make the whole column smaller.

PR #271, opened 2017-10-27
[NO MERGE] Cascade http client spike

mirabelrice:
This is a spike to play around with the cascade web service. This branch will serve as a reference for a future story to automate cascade deploy.
Story: https://trello.com/c/oBBHmLj3/471-spike-test-cascade-web-services-api-using-a-command-line-http-client


app/services/cascade_assets_client.rb
```diff
@@ -0,0 +1,86 @@
...
+        parentFolderPath: '_assets',
+        siteName: SITE_NAME,
+        metadataSetPath: '/Default',
+        shouldBePublished: 'false',
+        shouldBeIndexed: 'false',
+        data: encoded_image,
+        rewriteLinks: 'false',
+        maintainAbsoluteLinks: 'false'
+      }
+    }
```
mirabelrice: We will need to determine a way to represent different types of assets. For example, files, folders and images, etc. are all different types with different values in the request.

app/services/cascade_assets_client.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    message = {authentication: deploy_credentials, identifier: asset_identifier}
+    request(:read, message)
+  end
+
+  def write_asset_block(local_block)
+    # First get a reference to the remote asset block, and then update the xml with the local block
+    read_response = read_asset_block(ASSET_BLOCK_PATH)
+    return unless read_response && read_response.body[:read_response][:read_return][:success]
```
tatwell: I worry about unexpected silent failures here. Seems like we might want to log this or raise an exception.

PR #279, opened 2017-11-10
Change the name of "Happenings" to "News and Stories" in all our products and web environments.

tatwell:
Trello Card: https://trello.com/c/1CVI0zWI


tatwell:
@mandythomas I'm not sure exactly why those subdomains are enumerated in the code but that code wasn't changed with this update so it shouldn't affect functionality.

PR #292, opened 2017-12-05
[NO MERGE]Cascade http client spike

mirabelrice:
This is a spike story to play around with cascade web services. This branch will serve as a reference for a future story to automate cascade deploy.
Story: https://trello.com/c/wW23P2Go


mirabelrice:
This branch was closed before so it shows all the changes I made in the previous sprint as new commits. The main methods changed for this sprint were in [delete_assets method](https://github.com/chapmanu/cascade-assets/pull/292/files#diff-61862303ce5821187294dcfead0c4425R58) and [upload_assets method](https://github.com/chapmanu/cascade-assets/pull/292/files#diff-61862303ce5821187294dcfead0c4425R80)

app/services/cascade_assets_client.rb
```diff
@@ -0,0 +1,153 @@
...
+      asset: {
+        folder: {
+          name: File.basename(remote_path),
+          parentFolderPath: parent_dir_path(remote_path),
+          siteName: CASCADE_CONFIG['site_name'],
+          metadataSetPath: '/Website Folder'
+        }
+      }
+    }
+  end
```
mirabelrice: Because this was just a spike, I did not do this in the prettiest way. Eventually it would probably make sense to make an `Asset` class rather than having `file_asset`, `folder_asset`, etc..

lib/tasks/deploy.rake
```diff
@@ -0,0 +1,33 @@
+namespace :deploy do
+  namespace :cascade do
+    desc "Deploy cascade-assets to Cascade"
+    task all: :environment do
```
tatwell: How is this called from the command line?

mirabelrice: `rake deploy:cascade:all RAILS_ENV=development` will run all steps of the deploy process. I created each part of the deploy process as a separate task,  for example: `rake deploy:cascade:asset_block` for testing purposes. This can be changed in the final cap script if you want.

Issue #310, opened 2018-01-04
SiteImprove reports WCAG 2.0 violation: Element ID is not unique


tatwell:
Trello Cards: 

- https://trello.com/c/YxTEzOLv
- https://trello.com/c/tNeLAuXj

This issue started to be reported for almost every WWW page after we rolled out the new OmniNav. Here's the issue in SiteImprove (account access required):

- https://my2.siteimprove.com/Accessibility/10993/Pages/Issue?CheckId=33&CriterionChapter=4.1.1

Unfortunately, SiteImprove doesn't report which ID exactly is not unique. But we suspect it is associated with the OmniNav search form, which gets duplicated to accommodate desktop/mobile reflows.

tatwell:
I think I've identified the duplicate element. It's not quite where I expected.

Here is the markup for the desktop version of OmniNav search form (as seen on [Chapman home page](https://www.chapman.edu/)):

```
<div class="cu-search-box">
  <div id="___gcse_0" dir="ltr">
    <form class="gsc-search-box" accept-charset="utf-8">
      <table cellspacing="0" cellpadding="0" class="gsc-search-box">
        <tbody>
          <tr>
            <td class="gsc-input">
              <input autocomplete="off" type="text" size="10" class=" gsc-input " name="search" title="search" id="gsc-i-id1" dir="ltr" spellcheck="false" placeholder="Search" style="outline: none; background: url(&quot;https://www.google.com/cse/static/images/1x/googlelogo_lightgrey_46x16dp.png&quot;) left center no-repeat rgb(255, 255, 255); text-indent: 48px;">
              <input type="hidden" name="bgresponse" id="bgresponse">
            </td>
            <td class="gsc-search-button">
              <input type="button" value="Search" class="gsc-search-button" title="search">
            </td>
            <td class="gsc-clear-button">
              <div class="gsc-clear-button" title="clear results">&nbsp;</div>
            </td>
          </tr>
        </tbody>
      </table>
      <table cellspacing="0" cellpadding="0" class="gsc-branding">
        <tbody>
          <tr>
            <td class="gsc-branding-user-defined"></td>
            <td class="gsc-branding-text">
              <div class="gsc-branding-text">powered by</div>
            </td>
[...]

PR #311, opened 2018-01-09
Refactor OmniNav search module.

tatwell:
Trello Card: https://trello.com/c/tNeLAuXj
Issue #310


app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
+var GoogleCustomSearch = (function() {
```
tatwell: I've broken this out of the main `OmniNav2` module. Here's the before:

https://github.com/chapmanu/cascade-assets/blob/2758afd76755884feb074b9abdafbe394ad05b93/app/assets/javascripts/omni-nav-v2.js#L325

app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
...
+  };
+
+  /*
+   * This object represents the Search component (search form and results block). It wraps
+   * the Google element[0] and jQuery $elements representing the component in order to
+   * provide an API tailored to OmniNav.
+   *
+   * [0] https://developers.google.com/custom-search/docs/element#cse-element
+   */
+  var SearchComponent = function(containerId) {
```
tatwell: This replaces the `TwoColumnGCS` module:

https://github.com/chapmanu/cascade-assets/blob/2758afd76755884feb074b9abdafbe394ad05b93/app/assets/javascripts/omni-nav-v2.js#L384

I've renamed it to try to clarify its role and also tried to slim down the code a bit. For example, I eliminated as many module vars (under `// Internal Attrs`) as I could.

app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
...
+      // Bind only once: use a data flag to track binding.
+      var autoCompleteIsBound = $container.data('autoCompleteIsBound');
+      if ( autoCompleteIsBound ) return;
+
+      // This is required for selecting the Autocomplete block associated with
+      // SearchComponent. Since these are added post-loading by Google and are attached
+      // as absolute elements directly to the body, the parent cannot be deduced
+      // from the object itself. Checked Google API. No accessor. This is unfortunate.
+      var omniNavContext = ( containerId.indexOf("utility") >= 0 ) ? 'utility' : 'primary';
+      var autoCompleteSelector = contextSelectorMap[omniNavContext];
```
tatwell: I really don't like this. But I don't think there's a clear better option at this point. Especially as we still consider eliminating the two forms and have the `OmniNav2` module switch one back and forth depending on view dimensions.

app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
...
+      $container.data('autoCompleteIsBound', true);
+
+      // Bind callback: when autocomplete option is clicked, show results. Interval
+      // allows time for autoCompleteTable to be added to DOM.
+      var intervalId = setInterval(function() {
```
tatwell: This interval checking wasn't present in the original code but in testing on the local sample page the `$autoCompleteTable` would sometimes not be found. It's ugly but stopped the problem.

Have anyone noticed or received any reports of search results sometimes failing to load?

app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
...
+      $container.addClass('search-results-open');
+
+      // Bind callbacks.
+      lockScroll();
+      $(document).on('keyup', onSearchEsc);
+      $(document).on('click', onDocumentClick);
+
+      // Trigger a click on the GCS tab corresponding to the search type dropdown selection
+      // unfortunately GCS does not provide a callback for when search results
+      // are returned so must use timeout
```
tatwell: A common theme with GCS. This was brought over from original code. Makes me wonder if we're not using GCS in best way, but it looks like neither Mirabel or I found a better way in docs.

app/assets/javascripts/omni_nav/google-custom-search.js
```diff
@@ -0,0 +1,314 @@
...
+
+    var onSearchResultsResize = function() {
+      clearTimeout(resizeTimeoutId);
+      resizeTimeoutId = setTimeout(updateSearchResultsHeight, 500);
+    };
+
+    var bindAutocomplete = function() {
+      // Autocomplete map. This seems very fragile as it requires Google to use
+      // the same hardcorded selector each time. But it is how this was originally coded
+      // and so far has proved reliable so I am keeping it for now.
```
kaylaziegler: Yes, using the fragile selectors like this comment mentions is something we had to add when we started using two search boxes on the same page. If we only had one, we could simply do `$(".gsc-completion-container").on(...)` but now we have to figure out which is being clicked and the `gstl_50/51` was the one and only differentiating characteristic.

tatwell: It would be nice if the [google.search.cse.element object](https://developers.google.com/custom-search/docs/element#cse-element), `gcsElement`, offered access to the autocomplete element. That would make the most sense.

app/assets/javascripts/omni_nav/google-custom-search.js
kaylaziegler: @tatwell In regards to this comment and the one in the code, `Interval allows time for autoCompleteTable to be added to DOM`, that's what `setTimeout` was for here:
```
var onAutoCompleteClick = function() {
    setTimeout(showResults, 100);
};
```
Was that not working anymore? I never had an issue with it failing to load, while the timeout was there. I did test it briefly in the beginning just to be sure that it failed if the timeout was taken out.

tatwell: Yes, I think there are 2 potential race conditions here:

1. Bind `onAutoCompleteClick` callback to `$autoCompleteTable`: need to wait for Google to insert `$autoCompleteTable` into DOM.
1. When clicking `$autoCompleteTable`, waiting for something else to happen (Google loading results?) before showing results.

\#2 was already in the code. I tried removing both independently and got intermittent issues where an expected block would be blank or missing. So I think both are required. Using timeout is still prone to problems if the dependent event takes longer than the timeout delay.

As Mirabel noted, this is due to limitations in the Google API. It would be nice if they offered callbacks or events that could be used instead of this guesswork. I suppose it reflects the fact that they don't really want you extending their UI.

app/assets/javascripts/omni_nav/omni-nav-v2.js
```diff
@@ -193,15 +229,20 @@ var OmniNav2 = (function() {
...
       // If no name selector, look for ID selector.
-      if (!$target.length) {
+      if ( !$target.length ) {
         $target = $(idSelector);
       }
 
+      // Still no target? Return or we get an error below.
+      if ( !$target.length ) {
+        return;
+      }
```
tatwell: Maybe this isn't such a good idea. It's swallowing an error that maybe we should see. If this is getting called and `$target` isn't present, it seems like it would be an issue. Perhaps another race issue?

PR #312, opened 2018-01-17
Sync some data definitions with production; update some sample html

mandythomas:
None of these 5 changes will compile into a new cascade-assets; I was simply updating Data Definitions for the 2 and 3 Column Modulars to add some options in 2 dropdowns which I just made in production. The copies of the Data Defs in github were not current (even before my changes) with what's in production, so mostly just syncing them up. I also added 2 Data Defs that are in Cascade but not in github yet (the 2 for Shared Content blocks that we use for both Degrees and the Dept Contacts widgets). Last, I added an href to the Contact Footer in 1 Column sample html to match a change I made to a velocity format. Again, none of this rolls into a new deploy. They are all just sample code or saved copies of things in Cascade.


app/data_definitions/from_cascade/shared_content_degree_program.html
```diff
@@ -0,0 +1,108 @@
+<system-data-structure>
```
tatwell: For sake of consistency, these new files should probably have an `.xml` extension.

mandythomas: good catch. I've changed the extensions

PR #315, opened 2018-02-20
New OmniNav is causing accessibility errors on pages

### LINKED ISSUE #310: SiteImprove reports WCAG 2.0 violation: Element ID is not unique

kaylaziegler:
Story: https://trello.com/c/YxTEzOLv
Fixes #310 

This PR addresses a couple different SiteImprove and WAVE Accessibility errors/warnings.

- "Element ID is not unique."
  - For branded pages, the off-canvas has 2 close icons with the same ID that's used to trigger JS that closes the off-canvas menu. Changed it to a `class` attribute instead of `id`. Not ideal, but still works.
  - Some SVGs had redundant IDs that were unused and sometimes raised accessibility errors, so they've been removed.
- "Missing alt text."- Used JS to add alt text to Google image tag loaded in custom search box.
- "Missing or uninformative TITLE tag."- Adds text to SVG title tags in OmniNav off-canvas menu.
- "Non-distinguishable landmarks."- Adds `aria-label` to distinguish the two `<nav>` tags in the OmniNav.

See issue #310 for explanation about the original error that spawned this story. I didn't have to make any related code changes, but I did change it so that the Google custom search is only initialized once based on the window size when the user loads the page. That is just in case Google changes their code back, and to reduce the number of accessibility warnings/errors caused by Google's automatically generated search code, which we don't have control over.


mandythomas:
in buildGlobalNav macro in omni_nav_v2.vtl there's a param with typo. It says arial-label instead of aria-label. For the hamburger menu toggle icons, there's one with #buildToggleIcon("Find Information"). I wonder if it shouldn't be "Find Information By Audience" or something less generic. In several macros, eg buildOffCanvasNavCloseIcon macro you changed the js-close-off-canvas-nav id to a css class. Was that intentional? Two different purposes. In buildToggleIcon("Maps &amp; Directories") should we say 'and' instead of &amp; as it's for accessible Title of the icon svg, isn't it? Avoid accessibility possible issues reading &amp;

kaylaziegler:
@mandythomas I added a comment to the JS, and in the commit message, that explains why I had to change the ID to a class. I would prefer that it was an ID but it causes SiteImprove errors because we use it twice. Functionally, it makes sense for them to be exactly the same ID, but we don't have a choice here. We use one velocity macro to render the icon so we can't (and shouldn't) make them have different IDs. So the option I was left with was changing it to a class name instead of ID.

kaylaziegler:
@mandythomas Also, just so you're aware, I can change the title for toggle icon for the "Maps & Directories" dropdown so that it doesn't use the & symbol, but on branded pages with top-level pages that have & in the name, they will still show it (and not the word "and" written out). I don't think it causes issues, because I checked it in SiteImprove and WAVE.

Also looks like the & is one of the "safe" symbols for screen readers according to this article I found: https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/

mandythomas:
yeah, nothing to do with accessibility, but having duplicate ID on a page is invalid html markup so if you've fixed that, even though you didn't prefer it, it's actually the correct resolution ...if I'm understanding it all right. I'm guessing it still functions ok even though now it's a css class? An ID is for getElementById so you can't use that now if you needed. Last, you probably took care of this, but any styles based on it would have to change to using .classname now instead of #classname since it's not an ID now.

mandythomas:
I only mentioned the & ampersand because we've been getting some complaints about special symbols in certain places on the web and emails and I was looking to avoid that in case the ampersands in TITLE params didn't read out as the word "and" the way they do in visible text

kaylaziegler:
@mandythomas It does still function even as a CSS class. No stylesheet changes were necessary, the ID was just used to add the click event listener.

Made the other requested changes if you want to look at them (the typo, and changed the title for "Find Information", and removed ampersand from "Maps & Directions")

.cascade-code/Chapman.edu/_cascade/formats/modular/mastheads/side_panel_hero.vm
```diff
@@ -6,124 +6,127 @@
...
+            ## Featured Content Variables
+            #set ( $featuredCategory      = $slide.getChild('featuredPanel').getChild('category').value )
+            #set ( $featuredTitle         = $slide.getChild('featuredPanel').getChild('title').value )
+            #set ( $featuredSubtitle      = $slide.getChild('featuredPanel').getChild('subtitle').value )
+            #set ( $featuredBody          = $_SerializerTool.serialize($slide.getChild('featuredPanel').getChild('body'), true) )
+            #set ( $featuredLink          = $slide.getChild('featuredPanel').getChild('link') )
+            #set ( $featuredContentExists = "#if($featuredCategory != '' || $featuredTitle != '' || $featuredSubtitle != '' || $slide.getChild('featuredPanel').getChild('body').value.length() != 0 || $featuredLink.getChild('label').value != '')true#{else}false#end")
+            #set ( $featuredLinkExists    = $featuredLink.getChild('label').value != '' )
```
kaylaziegler: The majority of the stuff here is just indentation that I fixed. The only things I actually changed were this line (new variable called `$featuredLinkExists`) and lines 83 and 102 where I used it to test whether link data definitions were left empty by the user. For accessibility, links cannot be empty.

PR #318, opened 2018-02-28
On new OmniNav, fix mobile search box styling.

cbryant24:
fixed OmniNav search box alignment with `float: right` css property

story: https://trello.com/c/OCOmWbPB


kaylaziegler:
@cbryant24 Please remove all the changes to the files in the `build/` folder. I'm guessing you ran `rake build:omninav` per the instructions in the README. Looking at it now I just realized I need to update that. Technically, the builder only applies to the old OmniNav (v1) and your story is only for v2, so you don't need to run that. The Trello card will say very specifically if/when you ever have to make changes to the builder and run that rake task. 

Please read through the [OmniNav Wiki page](https://wimops.chapman.edu/wiki/OmniNav#OmniNav_Sites) for more details and let me know if you have questions.

PR #323, opened 2018-03-09
Fix to script for weather feed to get current temp

mandythomas:
In weather.js  the url to json feed of current temperature was using callback=? as url param, and now we need to use callback=orange per Nikos H. in Schmid College; also fixing links in footer blocks to link to forecast.chapman.edu with https as they installed an SSL certificate awhile back

### DIFF
--- a/app/assets/javascripts/cascade/weather.js
+++ b/app/assets/javascripts/cascade/weather.js
@@ -1,19 +1,16 @@
 $(function () {
-
-	/* Populate weather from json feed; default on page is 65 until modified by this routine
-    ------------------------------------------------------------------------------------------------*/
-	//sample data: {"weather":{"icon_path":"\/images\/icons\/weather\/2cloud_norain.png","temp_f":"66","temp_c":"19"}}
-	$.getJSON("https://forecast.chapman.edu/chapman/banner-json.php?callback=orange", function (data) {	
-
-        var iconPath = data.weather.icon_path,
-            tempF = data.weather.temp_f,
-            tempC = data.weather.temp_c,
-            url = "//www.chapman.edu",
-            $weather = $(".weather");
-
-        $weather.find("img").attr("src", url + iconPath);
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/cascade/weather.js
```diff
@@ -1,10 +1,9 @@
 $(function () {
 
-	/* Populate weather from jsonp feed
+	/* Populate weather from json feed; default on page is 65 until modified by this routine
     ------------------------------------------------------------------------------------------------*/
-
-    $.getJSON("//forecast.chapman.edu/chapman/banner-json.php?callback=?", function (data) {
-
+	//sample data: {"weather":{"icon_path":"\/images\/icons\/weather\/2cloud_norain.png","temp_f":"66","temp_c":"19"}}
+	$.getJSON("https://forecast.chapman.edu/chapman/banner-json.php?callback=orange", function (data) {
```
kaylaziegler: @mandythomas Can you change this file to 2-space tabs instead of 4 spaces?

app/assets/javascripts/cascade/weather.js
```diff
@@ -1,20 +1,16 @@
...
-
-        $weather.find("img").attr("src", url + iconPath);
-        $weather.find(".temp .f").html(tempF + "&deg; F");
-        $weather.find(".temp .c").html(tempC + "&deg; C");
-    });
-
+  /* Populate weather from json feed; default on page is 65 until modified by this routine
+  ------------------------------------------------------------------------------------------------*/
+  //sample data: {"weather":{"icon_path":"\/images\/icons\/weather\/2cloud_norain.png","temp_f":"66","temp_c":"19"}}
+  $.getJSON("https://forecast.chapman.edu/chapman/banner-json.php?callback=orange", function (data) {
```
mandythomas: @kaylaziegler I fixed the indenting of weather.js file. Since now it's harder to review with all lines showing as changed, I'm just, noting that the .getJSON line is the only real line with an important fix on it.

PR #326, opened 2018-03-12
Add site variable to omni nav format

mandythomas:
Added site://Chapman.edu which is a variable in Cascade which is uses to render the path for an internal link. Preview mode to assets within Cascade will still work, and when published the page will contain a fully qualified link to the site's defined url. This will allow it to publish omniNav markup to files on webfarm and links back to www.chapman.edu pages will work. Also, fixed link to /panthermail so fully qualified.


.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav_v2.vtl
```diff
@@ -801,4 +804,4 @@
 ## Main OmniNav HTML Block
 ##
 #setVars()
-#buildOmniNav()
+#buildOmniNav()
```
cbryant24: Just need a new empty line at the end of this file according to best practices.

mandythomas: @cbryant24 thx. added empty line to end

kaylaziegler:
@mandythomas What about the off canvas links? Those are set in `_config/global_nav.vm`, also in the .cascade-code directory in GitHub. Do you need to modify those as well?

mandythomas:
@kaylaziegler No we don't need to edit the _config files. As those urls are used in the main code, in an A HREF tag, that's where we put the site:// variable in for the link.

mandythomas:
@cbryant24 can you approve this now? I added the blank line you requested.

.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav_v2.vtl
```diff
@@ -307,7 +310,7 @@
             </a>
           </li>
           <li>
-            <a href="/panthermail/">
+            <a href="https://www.chapman.edu/panthermail/">
```
khafa100: @mandythomas Just wondering, but why is this one the only one that's not `site://Chapman.edu/`? At first I thought it was because it didn't have any variables, but then I noticed other `href`s in this file that didn't have variables either. So I just wanted to understand.

mandythomas: @kaylaziegler the /panthermail link is hardcoded with the domain because it's a 301 redirect setup in the IIS webservices on the WWW site and not a link to a real page in Cascade, And that same shortcut is not setup on other test or dev servers, it's only valid on WWW so we need to hardcode it in.

.cascade-code/Chapman.edu/_components/omni_nav_v2/omni_nav_v2.vtl
```diff
@@ -381,11 +384,15 @@
     <ul>
       ## First list item is always Overview
       <li>
-        <a class="icon $overviewIconClass" href="$overviewLink">Overview</a>
```
mandythomas: @kaylaziegler did some final review of the sample pages using this updated format and found that the 2 external links in the global nav dropdowns that are for athletics and events calendar were appending the site:// to the link when they shouldn't. I added an IF clause to test. Seems to work fine now as seen at https://dev-www.chapman.edu/pharmacy/support.aspx. Soooo, can you review these lines are 387+ and re-approve?

mandythomas: @kaylaziegler omg. good catch. I edited the file instead of copying the whole file I'd worked out on dev-cascade. thank you so much. glad we have code reviews.

PR #328, opened 2018-03-14
[Urgent] Fix OmniNav search function on Chapman.edu

kaylaziegler:
Story: https://trello.com/c/Am9ZQTZP

Just removed the javascript that added alt text to the image that google inserts in the custom search code. This was deemed the best solution because it doesn't introduce any errors, and it seems pretty likely that Google will change their class names again at some point so this was a bad way to do it.

I have not yet deployed to staging.

### DIFF
--- a/app/views/omni_nav_v2/_off_canvas_nav.html.erb
+++ b/app/views/omni_nav_v2/_off_canvas_nav.html.erb
@@ -23,6 +23,7 @@
       <div id="main-logo" class="toggle-logo">
         <a class="default off-logo" href="/" title="Chapman University Website Home Page">
           <!-- logo set as background image by class -->
+          Chapman University
         </a>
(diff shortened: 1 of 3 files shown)


app/views/omni_nav_v2/_off_canvas_nav.html.erb
```diff
@@ -23,6 +23,7 @@
       <div id="main-logo" class="toggle-logo">
         <a class="default off-logo" href="/" title="Chapman University Website Home Page">
           <!-- logo set as background image by class -->
+          Chapman University
```
kaylaziegler: I updated these 2 views because they were showing WAVE errors, even though the real velocity formats in Cascade have already been updated to fix them. The corresponding views we use to mimic those formats were not updated, and it's kind of distracting when I'm trying to look for real errors.

PR #330, opened 2018-03-15
Top Story on homepage not using right link fields for the Additonal Link

mandythomas:
The Additional Link field (ExtraLink) in the Top Story was pulling in a different
link field; Changed it to pull correct field(s), and test for internal vs
external; Followed conventions Tom had already established in this file previously when he refactored it.

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/Homepage_Blog_Stories
+++ b/.cascade-code/Chapman.edu/_cascade/formats/Homepage_Blog_Stories
@@ -28,7 +28,8 @@
 #set ( $topStoryCategoryText = '' )
 #set ( $topStoryCategoryLink = '' )
 #set ( $topStoryLink = '' )
-#set ( $topStoryLinkText = '' )
+#set ( $topStoryExtraLinkText = '' )
+#set ( $topStoryExtraLink = '' )
 #set ( $topStoryCtaLabel = '' )
 #set ( $topStoryExcerpt = '' )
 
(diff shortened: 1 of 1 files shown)


.cascade-code/Chapman.edu/_cascade/formats/Homepage_Blog_Stories
```diff
@@ -180,8 +183,15 @@
...
-        <p><a href="$topStoryLink">$topStoryLinkText »</a></p>
+      #if ( $topStoryExtraLinkText != "" )
+        #if ( $topStoryExtraExternalLink != '' )
+            #set ($topStoryExtraLink = $topStoryExtraExternalLink )
+        #elseif ( $topStoryExtraInternalLinkPath != '' && $topStoryExtraInternalLinkPath != '/' )
+            #set ($topStoryExtraLink = $topStoryExtraInternalLinkPath )
+        #else
+            #set ($topStoryExtraLink = '#')
+        #end
+        <p><a href="$topStoryExtraLink">$topStoryExtraLinkText »</a></p>
```
kaylaziegler: Do you want to use `${_EscapeTool.xml($topStoryExtraLinkText)` to allow special characters? I was able to break it by adding a & in the link text. Up to you, we could just tell the web coordinators not to use any special characters.

mandythomas: yes I will add that. The link text wasn't part of what I changed but should be using that function to avoid issues

mandythomas: added escape function as suggested.

Issue #305, opened 2017-12-20
OmniNav search form not working in Internet Explorer.


tatwell:
Trello Card: https://trello.com/c/5BNPbbsu

Sheryl notified Ross just now that search is not working for her in IE.

kaylaziegler:
This Stack Overflow question looks like the exact same issue we're experiencing: https://stackoverflow.com/questions/31119300/ie11-object-doesnt-support-property-or-method-includes-javascript-window

kaylaziegler:
The problematic line:
```
if($element[0].className.includes("utility") && !utilityAutocompleteBound)
```
Internet Explorer only works with indexOf to achieve what we want in this case. The purpose is to establish whether the user is submitting with the utility nav search or the primary nav search (since we currently have google custom search boxes in both places).

PR #335, opened 2018-04-04
As a user, I would like to see the emergency homepage banner updated to reflect new designs.

kaylaziegler:
Story: https://trello.com/c/lZ9jcZ0o

Also incorporates this story: https://trello.com/c/dlxo1Bnp

Currently deployed on staging: https://dev-www.chapman.edu/index.aspx


app/assets/javascripts/widgets/emergency-alert.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+    else {
+      $.ajax({
+        url: raveFeed,
+        type: 'GET',
+        success: function(data) {
+          $(data).find("item").each(function () {
+            var title = $(this).find("title").text();
+            // Looks for absence of the "no emergency" message
+            isEmergency = title.indexOf(noEmergencyMessage) < 0;
```
kaylaziegler: I really would prefer something like `.includes(noEmergencyMessage)` because it's more readable but we've encountered [browser issues](https://github.com/chapmanu/cascade-assets/issues/305) with that in the recent past.

app/assets/stylesheets/omni_nav/omni_nav_v2/omni-nav.scss
```diff
@@ -428,15 +428,15 @@ html.omni-nav-v2 {
             }
           }
 
+          .nav-icon-text {
```
kaylaziegler: This fixes the issue from this story: https://trello.com/c/dlxo1Bnp

The bug was preventing me from testing the "close" button on the emergency alert on mobile, so I fixed it.

app/views/widgets/home_page/_hero_panel.html
```diff
@@ -1,44 +1,28 @@
 <div id="heroWrapper">
```
kaylaziegler: The HTML in this file was outdated, which was making it hard to test my updates, and it's supposed to mimic the live site. So now it matches what's in production right now. Except it's just a static image instead of the video background.

app/assets/stylesheets/widgets/home_page/emergency_alert.scss
```diff
@@ -0,0 +1,105 @@
...
+        padding-bottom: 20px;
+        border-bottom: 1px solid white;
+
+        // Styles for potential WYSIWYG content (data def in Cascade)
+        p {
+          font-size: 16px;
+          line-height: 22px;
+        }
+
+        h1, h2, h3 {
```
mandythomas: not sure why we'd have the same size for h1, h2 and h3, or why we'd ever anticipate an h1 in emergency alert (would be invalid I think). Just curious

app/views/widgets/home_page/_emergency_alert.html
```diff
@@ -0,0 +1,20 @@
+<div class="emergency-alert">
+  <span class="close-alert-trigger">
+    <svg class="icon-close" version="1.1" xmlns="http://www.w3.org/2000/svg" width="13" height="13" viewBox="0 0 512 512">
+      <title>Close</title>
+      <path d="M446.627 110.627l-45.254-45.254-145.373 145.372-145.373-145.372-45.254 45.253 145.373 145.374-145.373 145.373 45.254 45.254 145.373-145.373 145.373 145.373 45.254-45.255-145.372-145.372z"></path>
+    </svg>
+  </span>
+  <div class="alert-text">
+    <svg id="icon-warning" xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32">
+      <title>warning</title>
```
khafa100: Should this also be capitalized to be consistent with the `title` above?

app/assets/stylesheets/widgets/home_page/emergency_alert.scss
kaylaziegler: This is kind of hard to explain. The design in the card allows for a title, which is automatically pulled from Rave and inserted into `<div class="alert-title">`. _Unless_ Ross (or whoever) enters a manual message in the data definition, which overrides everything from the feed. They would use the WYSIWYG for that, and all of the copy they put in the WYSIWYG gets put inside the general message div in the HTML and not the alert-title div, so it doesn't get the right styles, it won't look the way the design does (see trello). So in order to kind of force the design, I added these rules that just accommodate the different heading tags that Ross (or whoever) might use inside the WYSIWYG. I can change this CSS rule to just one heading level, but there's the possibility that someone will use a different heading level, which will end up looking pretty bad on the front end (I checked), just because of the default heading styles we have in place.

Up to you.

app/views/widgets/home_page/_emergency_alert.html
kaylaziegler: sure

app/assets/javascripts/widgets/emergency-alert.js
```diff
@@ -0,0 +1,76 @@
+var EmergencyAlert = (function() {
```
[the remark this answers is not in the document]
kaylaziegler: This is kind of the pattern we use for modules in javascript, see [flickr-picasa.js](https://github.com/chapmanu/cascade-assets/blob/master/app/assets/javascripts/cascade/level/flickr-picasa.js), [omni-nav-v2.js](https://github.com/chapmanu/cascade-assets/blob/development/app/assets/javascripts/omni_nav/omni-nav-v2.js), and [google-custom-search.js](https://github.com/chapmanu/cascade-assets/blob/development/app/assets/javascripts/omni_nav/google-custom-search.js) for a few examples. We use it to indicate modules in this case.

This is our best practices guide for that pattern: https://learn.jquery.com/code-organization/concepts/#the-module-pattern

PR #342, opened 2018-04-18
As an Alumni of Chapman, I would like to have "Alumni" in the global navigation.

kaylaziegler:
Story: https://trello.com/c/L3gmRexs

Tested the velocity format on staging just to make sure it works but I haven't been able to do a thorough test on staging yet because BREI is currently using it.

### DIFF
--- a/app/views/omni_nav_v2/_global_nav.html.erb
+++ b/app/views/omni_nav_v2/_global_nav.html.erb
@@ -45,6 +45,17 @@
           </ul>
         </div>
       </li>
+      <li class="primary-link">
+        <a>Alumni</a>
+        <div class="global-nav-dropdown">
+          <ul>
+            <li><a class="icon icon-file-text" href="/alumni/index">Overview</a></li>
+            <li><a class="icon icon-calendar4" href="/alumni/events/index.aspx">Events</a></li>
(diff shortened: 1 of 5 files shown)


app/views/omni_nav_v2/_global_nav.html.erb
```diff
@@ -45,6 +45,17 @@
           </ul>
         </div>
       </li>
+      <li class="primary-link">
+        <a>Alumni</a>
+        <div class="global-nav-dropdown">
+          <ul>
+            <li><a class="icon icon-file-text" href="/alumni/index">Overview</a></li>
```
mandythomas: this link to the overview in the global nav is not fully qualified page url with the .aspx in it like the others have in this section. Is just /alumni/index (line 52 of _global_nav_html.erb)

kaylaziegler: This is just the static sample we have in Cascade Assets to help us see what the page will look like. None of the links actually work. 

The velocity format we use to build the real omninav in cascade does have the fully qualified page URL, that works within the Cascade preview and on the published page. You can test it on the links in the product owner approval email I sent to Ross around 4:15

PR #379, opened 2018-08-14
Skip nav link

corejh:
You can see the skip nav link working on:

http://dev-www.chapman.edu/index.aspx
http://dev-www.chapman.edu/about/maps-directions/index.aspx

When you are on the page, without clicking anywhere on the page: hit tab. You should see a red button at the top right. It should respond to **enter** + clicks. It will take you to the #main, or the first h1 if #main doesn't exist. In the case it finds either #main/h1 and it's hidden, it will create a new div immediately after, and jump to that instead.  

Along with the skip nav link are several small improvements to cascade-assets, or to the javascript loading on the main page.


app/assets/javascripts/cascade/new-home.js
```diff
@@ -5,7 +5,7 @@
         if ( onHomePage ) {
             cu_window_manager.initialize();
             cu_hero_area.initialize();
-            cu_stories_area.initialize();
+            //cu_stories_area.initialize();
```
mandythomas: in javascripts/cascade/new-home.js I'm wondering why cu_stories_area is commented out now. Was there never any function that it called? In same script, there's several places where it replaces a path eg /_files with the fully qualified path such as http://www.chapman.edu path. I think now that SSL is mandatory on WWW site, those paths should be https not http. lines 422, 543, 547 etc.

cbryant24: I ran into the same issue when working on the footer story. The variable `cu_stories_area` is never created so this throws an "Uncaught TypeError: Cannot read property of undefined" error

mandythomas: Ok. I couldn't find it either. We definitely have an area on the page for Stories but it must not need any special Javascript. What about the changing of hardcoded http://www.chapman.edu to https instead?

corejh: I don't recall, also can't find that change in this PRs changed files. Can you point to exactly what you mean?

mandythomas: I checked the homepage on dev-www on IE, using https, and didn't get any warnings about mismatch so I guess it's ok as is.

PR #380, opened 2018-08-15
Update WWW Footer

cbryant24:
Story: https://trello.com/c/KZkTx9bi


app/assets/javascripts/cascade/footer.js
```diff
@@ -0,0 +1,27 @@
+$(function() {
+
+  skrollr.init({
+		forceHeight: false
+	});
```
cbryant24: We use skrollr library for responsiveness, in development it sets the height of the page on resize. Reading through the [documentation](https://github.com/Prinzhorn/skrollr) (which looks like it lost support as of 2014) to remove this behavior the config `forceHieght` needs to be set to false. @corejh

PR #345, opened 2018-05-02
Legacy mastheads

Sean-BREI:
Note: One Format was not updated to the most recent changes during my last Pull Request (1-Column Dropdown Feature) -- primary_column_helpers.velocity; that Format with those updates have been included in this Pull Request.

PR #391, opened 2018-09-18
Removing legacy mastheads mandy for sean

mandythomas:
this was an old pull request for a task given to Sean at Barkely REI. He'd removed the 3 old masthead options for 2/3 column modular templates. This was the option in the data def, and related formats and css and js for those mastheads. Mandy is trying to finish this up. Sidebar: the css for the placeholder and old branded masthead are being kept in a static file for webfarm.chapman.edu apps which still have some of those mastheads on pages (to be fixed later) but those css classes are being removed from our master.css as part of this PR


mandythomas:
merging Sean Stacho's pull request for removing the legacy masthead options from 2 and 3 Column templates and all the related css, js, velocity formats and data definitions. This was originally https://github.com/chapmanu/cascade-assets/pull/345. This new pull request is basically the same but with Dev merged into it and recreated as new request by Mandy. Related velocity and data defs in production Cascade are already updated. They did not depend on this PR going list into master.css (because this is the removal of things that are no longer used on any pages)

Issue #376, opened 2018-08-02
When tabbing through navigation, first enter doesn't trigger links


corejh:
Need to hit enter twice to use a link

corejh:
fixed by #397

Issue #373, opened 2018-08-01
When off canvas nav is not visible on screen, it's still tabbable


corejh:
If you tab through all the visible navigation elements, eventually you will hit the "hidden" off-canvas nav links, although they aren't shown on screen.

Expected behavior is they shouldn't be able to be targetted while hidden.

corejh:
fixed by #397

PR #397, opened 2018-10-13
Omninav improvements

corejh:
Fixes bug #376 and #373 

Also adds usability functionality to off-canvas nav. When user tabs (or uses down arrow) to tab past the last item in the menu, it will automatically close and continue to the next focusable item (chapman icon).

You can see fixes/improvements in effect here:
[Dev homepage](http://dev-www.chapman.edu)
[Dev about index (umbrella category)](http://dev-www.chapman.edu/about/index.aspx)


corejh:
@mandythomas Thanks for the heads up. I fast forwarded those sass files to use the same ones on development, so there shouldn't be any problems there. Files changed should strictly reflect omninav changes.

PR #410, opened 2018-12-04
Fix the links on www and other sources for news.chapman.edu

cbryant24:
Story: https://trello.com/c/BIarkMo0


cbryant24:
Oh I read @mandythomas comments backwards I'll check with her since they were h2s before so I changed them to h4s

cbryant24:
Should be updated now @corejh @mandythomas was referring to the links in news.chapman.edu

PR #414, opened 2018-12-07
Resolve utility nav contrast errors

nicholasnadel:
https://trello.com/c/d9syslSb

Demo via Wave (non-visual change) at https://dev-www.chapman.edu/nick/off-canvas-utility-contrast-fix.aspx


nicholasnadel:
@corejh I've adjusted VSCode's formatter indent spacing; Prettier is opinionated in terms of whitespace around sass selectors and I haven't found a quick way to override it (presumably related to https://stackoverflow.com/questions/18060132/what-is-ul-means-in-sass)

PR #422, opened 2019-01-24
Added new icomoon icons for ad_landing/feature_points_widget + Clean up

corejh:
Deleted the extra data definitions folder. There was already one inside app/data_definitions/from_cascade, although in the future we'll probably want to separate that from the rails app.


corejh:
@cbryant24 @mandythomas could I get a code review

Issue #429, opened 2019-02-08
Can tab to off-canvas nav


corejh:
duplicate issue of #397. Something happened and bug was reintroduced

PR #430, opened 2019-02-08
Fix omninav tab issues

corejh:
fix issues #428 and #429

also small qol changes such as adding a focus outline for off-canvas nav X button and minor clean up


corejh:
Sorry forgot to add: You can see changes here: https://dev-www.chapman.edu/test-section/joey-test/2col-test-page.aspx

PR #450, opened 2019-02-20
Accessibility Siteimprove 2.4.4

nicholasnadel:
https://trello.com/c/bNrtZRm5

Address Siteimproves "Link text used for multiple different destinations". 

Some sections use the same text to describe different links, eg `"Alumni" > "Get Involved" (https://dev-www.chapman.edu/alumni/get-involved/index.aspx)` and `"Support" > "Get Involved" (https://dev-www.chapman.edu/support-chapman/get-involved.aspx)` .  I've applied aria-labels where appropriate.

You can demo with the Siteimprove extension at https://dev-www.chapman.edu/test-section/nick-test/index1.aspx and compare against the former issues at https://my2.siteimprove.com/Inspector/10993/Accessibility/Page?pageId=31903631364&impmd=0D6974F8A4BBF3C0E325AAA2E369A7D2#/Criterion/2.4.4/Check/39

Note that Siteimprove 2.4.4 will now always display "Does the "aria-label" accurately describe the link?" ... "This warning will appear for all of your 'aria-label' on links since a manual review is needed."

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/Homepage UG Admission Section
+++ b/.cascade-code/Chapman.edu/_cascade/formats/Homepage UG Admission Section
@@ -65,20 +65,20 @@
             </ul>
 
             <div class="actions">
-                <a class="button white smc-cta" data-cta-id="homepage-ug-adm" data--100-bottom="opacity:1" data-bottom="opacity:0" href="https://www.chapman.edu/discover/index.html">Discover Chapman</a>
+                <a class="button white smc-cta" data-cta-id="homepage-ug-adm" data--100-bottom="opacity:1" data-bottom="opacity:0" href="site://Chapman.edu/discover/index">Discover Chapman</a>
                 
                 ### popup form not used: <a class="button red request admissionCTA_start" data--100-bottom="opacity:1" data-bottom="opacity:0" href="javascript:void();">Request Information</a>
                 #if ($leftButtonTrackID != "")
-                    <a class="button white smc-cta" data-cta-id="$leftButtonTrackID" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$leftButtonLink">$leftButtonLabel</a>
+                    <a class="button white smc-cta" data-cta-id="$leftButtonTrackID" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$leftButtonLink" aria-label="$leftButtonLabel">$leftButtonLabel</a>
                 #else
-                    <a class="button white" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$leftButtonLink">$leftButtonLabel</a>
+                    <a class="button white" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$leftButtonLink" aria-label="$leftButtonLabel">$leftButtonLabel</a>
                 #end
                 #if ($rightButtonTrackID != "")
-                    <a class="button white smc-cta" data-cta-id="$rightButtonTrackID" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$rightButtonLink">$rightButtonLabel</a>
+                    <a class="button white smc-cta" data-cta-id="$rightButtonTrackID" data--100-bottom="opacity:1" data-bottom="opacity:0" href='$rightButtonLink' aria-label="$rightButtonLabel">$rightButtonLabel</a>
                 #else
-                    <a class="button white" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$rightButtonLink">$rightButtonLabel</a>
+                    <a class="button white" data--100-bottom="opacity:1" data-bottom="opacity:0" href="$rightButtonLink" aria-label='{$rightButtonLabel}'>$rightButtonLabel</a>
                 #end
 
             </div>
         </div>
-</section>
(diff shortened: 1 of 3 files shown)


nicholasnadel:
> in omni_nav_v2.vtl
> 
> don't think we need lines 446 and 447 afterall. We aren't adding any aria-label like the others in this if/else and since we aren't doing the 'shortcut' url for it anymore it should just work as is and doesn't need to be here
> 
> don't think we need lines 448 and 449 because the link text is Discover Chapman so we don't need an aria-label that says identical thing. For same reason, don't need the aria-label seen on line 648 (though still need rest of that line)
> 
> there's a trailing space in A tag on line 481

@mandythomas I've committed the changes and added to dev. You can demo again at https://dev-www.chapman.edu/test-section/nick-test/index1.aspx . Note that the `aria-label="Discover Chapman" was to workaround the hardcoded URL in `Chapman.edu/_cascade/formats/Homepage UG Admission section`, so Siteimprove will again throw the 2.4.4 error for that specific button. Presumably this won't be an issue in production

mandythomas:
@nicholasnadel The link in the UG Admission section (the hardcoded link to Discover on the live site) was not done correctly. It like all the others should really be site://Chapman.edu/discover/index. I fixed that format on DEV. Let me know if I should move that to prod or let it be part of your github pull request. It's /_cascade/formats/Homepage UG Admission section.  That should eliminate the need for that aria-label and if/then line for Discover Chapman in the omniNav, right?

nicholasnadel:
> @nicholasnadel The link in the UG Admission section (the hardcoded link to Discover on the live site) was not done correctly. It like all the others should really be site://Chapman.edu/discover/index. I fixed that format on DEV. Let me know if I should move that to prod or let it be part of your github pull request. It's /_cascade/formats/Homepage UG Admission section. That should eliminate the need for that aria-label and if/then line for Discover Chapman in the omniNav, right?

Thanks and yes I can confirm the alert no longer appears at https://dev-www.chapman.edu/test-section/nick-test/index1.aspx. I'll add  `/_cascade/formats/Homepage UG Admission section` to this PR to be kosher

nicholasnadel:
> @nicholasnadel for each of the links in that Homepage UG Admission section format file, you've added an unnecessary aria-label. You're setting each button with an identical value in aria-label as the link's default text. You don't need an aria-label at all (unless you're specifiying a different or missing value) which we aren't. Here's a snippet of one: ...aria-label="$leftButtonLabel">$leftButtonLabel

@mandythomas I have not edited this file; I copied and pasted your changes. I can remove this if you would like

mandythomas:
odd. those aria-labels appear in that format as new changes per the pull request. I also looked on Dev-Cascade and they are not in there in older versions but Feb 15 in one of your edits they appear. It's probably a matter of those not being the same in github as they are in one or both Cascade instances. Yes, please remove the extraneous aria-labels in your new version of this format (and in Dev Cascade) so we are fixing these things as we find them (and consistent in all our environments). Makes me wonder what else was different between github, dev and prod cascade for this file.

nicholasnadel:
> odd. those aria-labels appear in that format as new changes per the pull request. I also looked on Dev-Cascade and they are not in there in older versions but Feb 15 in one of your edits they appear. It's probably a matter of those not being the same in github as they are in one or both Cascade instances. Yes, please remove the extraneous aria-labels in your new version of this format (and in Dev Cascade) so we are fixing these things as we find them (and consistent in all our environments). Makes me wonder what else was different between github, dev and prod cascade for this file.

@mandythomas I'm not sure, I added this file to version control 12 days ago and it wasn't on GH prior. I've removed the aria-labels from `Chapman.edu/_cascade/formats/Homepage UG Admission Section`

demo: https://dev-www.chapman.edu/test-section/nick-test/index1.aspx

mandythomas:
@nicholasnadel not sure if you'd made all your changes but looked tonight and found these outstanding issues:
"Homepage UG Admission" -- there is still an aria-label that is identical to the link's text so do not need the aria-label at all. It's on line 72 eg ....aria-label="$leftButtonLabel">$leftButtonLabel</a>

"omni_nav_v2.vtl" -- on  the link for /about/connect/index, since it is now a regular internal link (and now corrected in the _config/global_nav) there's no need to hardcode it in to this format, it can just use the value from $dropdownLink.link like the lines above it do. So line 445 could just be href="$dropdownLink.link" like line 443 (keep aria-label though)

nicholasnadel:
> @nicholasnadel not sure if you'd made all your changes but looked tonight and found these outstanding issues:
> "Homepage UG Admission" -- there is still an aria-label that is identical to the link's text so do not need the aria-label at all. It's on line 72 eg ....aria-label="$leftButtonLabel">$leftButtonLabel
> 
> "omni_nav_v2.vtl" -- on the link for /about/connect/index, since it is now a regular internal link (and now corrected in the _config/global_nav) there's no need to hardcode it in to this format, it can just use the value from $dropdownLink.link like the lines above it do. So line 445 could just be href="$dropdownLink.link" like line 443 (keep aria-label though)

Thanks Mandy-- I've removed these outstanding aria-labels. Hope you were not working late last night on my account! If approved, I'll close this and issue a new Pull Request for you, Joey, and Chris to approve

PR #456, opened 2019-03-11
WWW, Law footer is broken on mobile view

cbryant24:
story: https://trello.com/c/6kz40g3R

### DIFF
--- a/app/assets/stylesheets/base/animations.scss
+++ b/app/assets/stylesheets/base/animations.scss
@@ -18,7 +18,7 @@
     }
      100% {
       opacity:0;
-      height: 0em;
+      height: 0;
       transform:scale(0.9) rotateX(-60deg);
     }
 }
--- a/app/assets/stylesheets/base/animations.scss
+++ b/app/assets/stylesheets/base/animations.scss
@@ -1,13 +1,11 @@
 @keyframes footerLinksIn {
   0% {
     opacity: 0;
-    height: 0;
     transform:scale(0.9) rotateX(-60deg);
     transform-origin: 50% 0;
   }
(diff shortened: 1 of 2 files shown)


mandythomas:
@cbryant24 the pull request itself seems fine but when I looked at this page https://dev-www.chapman.edu/law/index.aspx that I assume shows the revised Law Footer, and scale browser down to mobile size, the icons in the "social" section (when expanded) are really close, almost overlapping, the top of the Quick Links section. The other sections seem to have more margin. Also since the files in this pull request are not just used in the Law Footer, they are used in regular page's footer also, did you apply this cascade-asset version to other pages to make sure that what you did for Law didn't mess up non-Law pages?

corejh:
@mandythomas could you add a screenshot of what you're seeing?

cbryant24:
@mandythomas I'm not seeing the issue with social icons you mention here's how it looks when I use the mobile view. I also published to homepage as well to display the footer still has the same look with these changes
image
image

mandythomas:
@cbryant24 here is a screenshot of what I was seeing on a PC in Firefox after I viewed it in full screen then dragged corners down to mimic mobile width. However I played around with the browser width and refreshed and then I could never replicate it again. It had happened this afternoon, and tonight, but now shows fine every time. I'll approve the pull request though and you can decide if we worry about this. If you can't replicate it either then I don't know...it was a fluke?
law-footer-screenshot-mobile

cbryant24:
@mandythomas I did make a change yesterday to the margin and padding due to an issue I noticed on Firefox with the animation so I believed that addressed the issue on firefox for PC as well

PR #459, opened 2019-03-13
News & Events tab widget fix

cbryant24:


mandythomas:
@cbryant24 In app/assets/javascripts/cascade/news-events.js you left in the comment I had added in 2017 saying I didn't think this JS was used. Since you discovered it is, and we have 2 called news-events.js I think you should do the following: Remove my comment and add comments into both describing what they do to distinguish one from the other. You also left in some lines with //commenting-out code and I thought when we checked things into github we didn't want to leave old commented-out lines in the file

mandythomas:
@cbryant24 The data definitions that are supposed to be a match of what's in Cascade are in app\data_definitions\from_cascade\ folder. Looks like you changed app/data_definitions/one_column.rb which I think was Tom Atwell's side project (trying to replicate their functionality in the local environment). I don't believe we actually use those in the local env, or copy those to Cascade. Should you have made the same change to the one in the from_cascade folder (first comparing it to prod or dev Cascade as I'm not sure if it's current as is).

mandythomas:
@cbryant24 I haven't heard of the social04.chapman.edu server before. That's setup outside the firewall, and has our normal backups and security patches applied like all servers should, is that right?

cbryant24:
@mandythomas I'll add the comments to the other news-events.js and removed that commented out line. There was other commented out code that was previously there as well that I will also remove. My issue is I'm not sure where this js file is being utilized, would you say it would be safe to remove this file from cascade completely? When I look through it it looks like an older version of what the other news-event.js file is doing.  

That one_column.rb file that I changed is used purely for development in the rails app. There's no need to make any changes in cascade. I changed that file due to an error it was causing when trying to develop locally. 

social04.chapman.edu is the same server we use for the social.chapman.edu node application for streaming so it should have all the same security, right? This is another node application that is using the same configuration but using a different port.  Should I check with Chris on the server team?

mandythomas:
@cbryant24 well now I'm confused. I had at one time put that comment in app/assets/javascripts/cascade/news-events.js saying I suspected it wasn't used. Then I *thought* you'd said you did end up having to modify it in order for the links at the bottom of the 3 items to resolve to the right urls. So if that is the case why are you suggesting deleting it (or am I misunderstanding something)?

cbryant24:
@mandythomas it is included in the javascript files and due to the `$(function () {})` function it is run immediately and replaces the href for the `.allEvents` class link. What I think was happening is this file ran before the other news-events.js file since the api call to Yahoo to parse the xml data and return JSON data took longer the `.allEvents` href link was changed after this file ran but now that the api call is faster for xml parsing this file runs after and caused the `.allEvents` class href to be updated to `/events/calendar.aspx` it never caused an issue before so I think it was left untouched.

mandythomas:
@cbryant24 in the spirit of cleaning up... yes I think we should remove that js file that you don't think is used anymore. I'd recommending deleting it completely from github cascade-assets repo (but keep a local copy somewhere on your computer just in case we're wrong). Compile new master.js and test on dev-cascade and dev-www pages. If it still works then we know and we got rid of extrameous code.

If social04 is a live server used for other things, and been around for a while it is likely fine. If you want to check with Chris Bolduc you can but I'm not as worried. I had thought maybe it was a new server or private or something

cbryant24:
@mandythomas I verified with Chris and he stated that social04.chapman.edu points to insdv-mkt-ap01.chapman.edu so that server should be fine. I tested a few widgets with that file removed and it didn't cause any issues. I'll remove that file and push those changes.

mandythomas:
@cbryant24 chris, did you remove that extraneous js for news/events and fix the comments so we are sure which files are actually used, and for what?

cbryant24:
@mandythomas yes the news/events js file `news-events.js` file in `app/assets/javascripts/cascade` should now be removed

PR #462, opened 2019-03-20
Remove links to blogs landing page

mandythomas:
find links to blogs.chapman.edu landing page and replace with news.chapman.edu. Most changes were really just to the views and other static files for localhost and not to anything used on the live site in master.js


mandythomas:
This pull request is to swap out links to blogs.chapman.edu for news.chapman.edu, where applicable. Most of these changes are to views used only in localhost testing.  I updated the build/omninav/blogs/omni-nav.php, build/omninav/inside/omni-nav.html for consistency, but don't intend to deploy those to blogs or inside servers. On chapman_stories_feed.js only change was to default value for the Button's url. Changing the baseBlogUrl from blogs to news made the widget not pull any of right stories, so will make separate ticket to address that.

PR #464, opened 2019-04-05
Feature - back to top button

nicholasnadel:
story: https://trello.com/c/UOGowamS

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/blocks/html/global_footer.html
+++ b/.cascade-code/Chapman.edu/_cascade/blocks/html/global_footer.html
@@ -1,102 +1,236 @@
 <!--
 global_footer.html
 Master: Chapman.edu/_cascade/blocks/html/Global Footer
---><footer class="footer">
-  <div class="inner-container">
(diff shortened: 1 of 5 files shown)


mandythomas:
@nicholasnadel there is a LAW version of the global footer that probably needs the same thing done to it. In Cascade instead of Global Footer block they use Law Footer block. It likely shares the same CSS though. This page https://dev-www.chapman.edu/law/index.aspx, or any other in /law folder would be good for you to test on

nicholasnadel:
> @nicholasnadel there is a LAW version of the global footer that probably needs the same thing done to it. In Cascade instead of Global Footer block they use Law Footer block. It likely shares the same CSS though. This page https://dev-www.chapman.edu/law/indhex.aspx, or any other in /law folder would be good for you to test on

Thanks @mandythomas. I've added the code which you can inspect above and demo at https://dev-www.chapman.edu/test-section/nick-test/sprint-4-10-19/law-back-to-top.aspx

nicholasnadel:
@corejh please take a look when you're able

PR #470, opened 2019-04-22
Hotfix - WYSIWYG Anchor Point

nicholasnadel:
Screen Shot 2019-04-22 at 5 52 32 AM

### DIFF
--- a/app/assets/javascripts/cascade/campus-services-anchor-point-readjustment.js
+++ b/app/assets/javascripts/cascade/campus-services-anchor-point-readjustment.js
@@ -1,7 +1,6 @@
 // Accounts for Omninav height when scrolling to anchor links in WYSIWYG https://trello.com/c/ZCjBDv9e
 $(function () {
-  var url = location.pathname;
-  if (url.indexOf('campus-services/marketing-communication/guidelines-and-resources/writing/') > 0) {
+  if ($('.wysiwyg-widget').length > 0) {
     $('.wysiwyg-widget a[href*=\\#]:not([href=\\#])').click(function () {
       if (
         location.pathname.replace(/^\//, '') == this.pathname.replace(/^\//, '') &&
--- a/app/assets/javascripts/cascade/campus-services-anchor-point-readjustment.js
+++ b/app/assets/javascripts/cascade/campus-services-anchor-point-readjustment.js
@@ -1,14 +1,8 @@
 // Accounts for Omninav height when scrolling to anchor links in WYSIWYG https://trello.com/c/ZCjBDv9e
 $(function () {
-  if ($('.wysiwyg-widget').length > 0) {
-    // Remove spaces in anchor link urls
-    $('a').each(function (removeSpace) {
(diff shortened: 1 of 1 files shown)


nicholasnadel:
> First thanks for jumping on this. Second, the fix resolves the sitewide broken link issue so from that perspective I approve the request. I put your cascade-assets block on a test page with my original 2 links that were failing and is ok now. However this hotfix is to name a specific page (campus-services/marketing-communication/guidelines-and-resources/writing/) to which the function applies which seems like bad practice because a) the page could move and b) it doesn't fix any other pages that might do whatever it is that this page did that required a fix. What is it that's not right on that page that requires you to modify the url upon page load? I think on this particular page, last month, I'd manually gone thru and changed the named anchors so they didn't contain blanks. Does that negate the need for your function? If not, let's push this thru and I'll do a sitewide publish of the site over lunch-break so maybe it impacts less of our campus users, then come up with a better longterm solution to that A-Z page's issue with the anchor links.

Anytime, @Mandy. Are there similar WYSIWYGS we can test? I've removed the remove_space() function entirely, so it _shouldn't_ cause issues elsewhere. I just thought I should be extra safe and narrow it to this specific URL for the hotfix.

mandythomas:
@nicholasnadel I'm not sure what you're asking. Almost every page on the website has a section where someone used the rich text editor to enter content. What specifically can I help you find? More importantly what is it about that A-Z page that required your original function to modify the links to their named anchors. Is it just that they contained blanks in the names or something else? I'd like to understand better what you were trying to fix when you first created this function. 

Not sure this helps but this page has named anchors if that's what you were asking https://www.chapman.edu/campus-services/information-systems/security/faq.aspx (though none have blanks in the anchor name)

nicholasnadel:
> . Almost every page on the website has a section where someone used the rich text editor to enter cont

> .  More importantly what is it about that A-Z page that required your original function to modify the links to their named anchors.

I was trying to prevent future issues where _anchor_ links contained **spaces**, since they don't work properly otherwise. In other words, the user had entered `<a href="#example link with literal whitespace in it>` and smooth scrolling did not work. I imagine that going forward, the user can be trained or instructed to use proper syntax without whitespace... 

I'll remove the conditional URL path part of the statement to test https://www.chapman.edu/campus-services/information-systems/security/faq.aspx.

nicholasnadel:
@mandythomas pushed the changes to DEV cascade-assets-nick. 
 
All appears well at https://dev-www.chapman.edu/test-section/nick-test/anchor-hotfix-information-systems-faq.aspx and https://dev-www.chapman.edu/campus-services/marketing-communication/guidelines-and-resources/writing/index-hotfix.aspx

mandythomas:
@nicholasnadel I scanned the published pages on WWW to see which had named anchors, and of those, which had embedded blanks in the anchor name. Only this page https://www.chapman.edu/law/admission/admissions-policies.aspx. I will address it with them. Let's consider it a training issue -- authors need to create anchors without whitespace in the names. Ready for you to push your changes to that JS and cascade-assets to production Cascade

PR #473, opened 2019-04-26
removing MyWindow and Discover Chapman links from omniNav and hamburger menus

mandythomas:


mandythomas:
merging my changes to remove myWindow and Discover links from menus; and Nick's changes going into master too (already deployed those to production per email from him)

PR #478, opened 2019-05-07
move leftover CSS causing global <label> conflicts

nicholasnadel:
Resolves global styling issue with `<label>` related to the new upgrade-alert notice. 

before: https://www.chapman.edu/website-feedback-form.aspx
after: https://dev-www.chapman.edu/website-feedback-form.aspx

### DIFF
--- a/app/assets/stylesheets/themes/home.scss
+++ b/app/assets/stylesheets/themes/home.scss
@@ -487,32 +487,3 @@ h6,
 /* End of file */
 
 
-
-label.upgrade-notice {
-  float: right;
-}
-
-#upgrade-browser {
-  background-color: #FFFFBD;
-  font-size: 24px;
-  font-weight: bold;
-
-}
-
-#upgrade-hide {
-  display: none;
-  transition: 1s ease-in-out;
-}
-
-label {
-  cursor: pointer;
-  text-decoration: underline;
-}
-
-#upgrade-hide:checked~#upgrade-browser {
-  display: none;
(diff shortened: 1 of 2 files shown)


mandythomas:
@nicholasnadel isn't the label text-decoration: underline on line 73 in the upgrade_alert.scss still applied to the whole page and not just the Alert div? Not related to this specific fix, but is your Upgrade Alert only going to be in effect on the homepage and not the other pages on the site? It seems so since your classes are in the home-page folder's CSS

nicholasnadel:
> @nicholasnadel isn't the label text-decoration: underline on line 73 in the upgrade_alert.scss still applied to the whole page and not just the Alert div? Not related to this specific fix, but is your Upgrade Alert only going to be in effect on the homepage and not the other pages on the site? It seems so since your classes are in the home-page folder's CSS

It's hidden in the comparison snippet, but if you view the [whole SCSS file](https://github.com/chapmanu/cascade-assets/blob/a08c39c0a50efaa53452acb45a0c59a826f817c9/app/assets/stylesheets/widgets/home_page/upgrade_alert.scss), it's nested and will be prefixed with `#upgrade-browser {`:

```
#upgrade-browser {
  #upgrade-hide {
    display: none;
    -webkit-transition: 1s ease-in-out;
    transition: 1s ease-in-out;
  }

  label {
    cursor: pointer;
    z-index: 2;
  }
```

Thanks for looking out

PR #479, opened 2019-05-08
Hotfix - min supported browsers

nicholasnadel:
This is a follow-up to https://trello.com/c/5Iw1jIh7 

Essentially, the alert was hidden, but blocking Omninav links (`position: absolute` + `z-index`). That code was removed, so it's not currently an issue. This resolves that problem and maintains functionality. 

I've also added the option to 'Do not display this message again' via local storage, which will save the pref until cookies/cache are cleared. 

I've tested, but if anyone has time I'd really appreciate additional eyes/ stringent testing before pushing to WWW. 

example: https://live.browserstack.com/dashboard#os=Windows&os_version=XP&browser=IE&browser_version=8.0&zoom_to_fit=true&full_screen=true&resolution=responsive-mode&url=http%3A%2F%2Flocalhost%3A5000%2Fhome_page%2Fsample&speed=1&host_ports=google.com%2C80%2C0

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/omni_nav_v2.vtl
+++ b/.cascade-code/Chapman.edu/_cascade/formats/omni_nav_v2.vtl
@@ -847,17 +847,18 @@
 #setVars()
 #buildOmniNav()
 
+
 ## Browser Upgrade Alert 
 <div class="upgrade-browser-wrapper">
     <input type="checkbox" id="upgrade-hide"/>
     <label for="upgrade-hide" class="upgrade-notice dismiss">
         <a class="fa fa-window-close no-underline" aria-hidden="true"></a>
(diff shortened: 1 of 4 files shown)


nicholasnadel:
> Unsure why but when testing in browser stack IE10 "do not show again" is not working
@corejh I think cascade-assets.xml had been overridden... I've got it set to a custom one and it's working for me on Browserstack/Windows 8/ie10, if you wouldn't mind trying ie10 again 

http://recordit.co/pi7DNJCzAu

corejh:
Okay I think I see what's happening. I was testing on dev-www.chapman.edu , I thought the code was all there since the notice appears. When I try on your test page it seems to work, but it's all sorts of broken for the homepage

PR #483, opened 2019-05-22
WWW/Newsroom, Homepage Story Automation

cbryant24:
Story: https://trello.com/c/fKbgQfMo


cbryant24:
@corejh whoops good catch didn't notice that. Should be fixed now.

mandythomas:
@cbryant24 why are you creating a new folder named Manage Site. I think the homepage formats should go in cascade-assets\.cascade-code\Chapman.edu\_cascade\formats\home\ in github. do you agree?  I've also always wanted to move the actual formats in Cascade (that are all related to the Homepage) into a home or homepage subfolder too. Could you do that so we are doing some much needed cleanup and reorganization as we go? Can you ask @corejh if that can be part of this (if not much more effort)

cbryant24:
@mandythomas I wanted to include that file, but I didn't know that `cascade-assets.cascade-code\Chapman.edu_cascade\formats\home\` referred to homepage formats. I like that idea of creating  a subfolder called homepage. @corejh Are you good with that creating a subfolder for the homepage? @mandythomas  were you thinking `cascade-assets.cascade-code\Chapman.edu\_cascade\homepage`

corejh:
@mandythomas @cbryant24 what exactly would we be moving into this folder? does it make sense, and are there any gains for doing this?

PR #488, opened 2019-05-31
appears to resolves infinite loading issue without conflicts at https…

nicholasnadel:
Follow-up to #485, where the `.news` selector was clashing with https://www.chapman.edu/events/chapman-celebrates/index.aspx and preventing this and all pages in the directory from loading.

story: https://trello.com/c/TStyuPBh

This simply wraps all of the code in `  if ($(".news-events-widget").length) {
`  instead of prepending `.news-events-widget` line-by-line. 

This can be tested against the "before" page by pasting in the console, as we don't have this page on dev. The "after" demo is a working example on the left-col page.

before: https://www.chapman.edu/events/chapman-celebrates/index.aspx
after: https://dev-www.chapman.edu/test-section/nick-test/sprint-6-6-2019/left-col-featured-news-events3.aspx

### DIFF
--- a/app/assets/javascripts/cascade/level/news-events.js
+++ b/app/assets/javascripts/cascade/level/news-events.js
@@ -41,10 +41,11 @@ $(function () {
       return null;
     }
   };
-  
-  /* Populate news from Wordpress RSS feed (converted to JSON with YQL)
-  ------------------------------------------------------------------------------------------------*/
-  if ($(".news-events-widget.news").length) {
+
+  if ($(".news-events-widget").length) {
+    /* Populate news from Wordpress RSS feed (converted to JSON with YQL)
+    ------------------------------------------------------------------------------------------------*/
+
     //default is NewsAndStories:
     var newsFeedUrl = "https://www.chapman.edu/getFeed.ashx?name=newsNewsAndStories",
       newsYqlUrl = function () {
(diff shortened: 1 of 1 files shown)


corejh:
Not sure if this specifically would ever be a problem, but wouldn't this fix not work for any news-events-widget placed inside /news/ or /events/?

nicholasnadel:
> Not sure if this specifically would ever be a problem, but wouldn't this fix not work for any news-events-widget placed inside /news/ or /events/?

I might be mistaken, but now that the widget contains the class ".news-events-widget", I think it should still run:

https://dev-www.chapman.edu/test-section/nick-test/sprint-6-6-2019/news/left-col-featured-news-events3.aspx

https://dev-www.chapman.edu/test-section/nick-test/sprint-6-6-2019/events/left-col-featured-news-events3.aspx

nicholasnadel:
> if you need to push this thru go ahead otherwise I want to try different situations of the widget on DEV on pages. for example, your samples are only in left col not primary and only if they choose to show both news and events. they could show News in left and 2nd widget in primary with Events or Events & News. does it work in all those...I will create some pages with those combos and make sure. But if we need t push this because it's broken sitewide, don't want for my approval. The above situationsare probably seldom seen

More than happy to wait. `/_cascade/blocks/html/nick-assets/modular_assets/news-events-loading` contains these changes if you want to use that for convenience.

nicholasnadel:
@mandythomas just nagging to see if you've had a chance to test

PR #505, opened 2019-07-09
Feature - global smooth scrolling with HTML & CSS only

nicholasnadel:
story: https://trello.com/c/ODSVPxGf

- Comments out global.js (happy to remove if that makes more sense) and implements smooth scrolling + offset with pure html and css
- The commits on Jun 12 are related to a hotfix for the Image Slider Widget on https://www.chapman.edu/industry-partners/index.aspx


nicholasnadel:
@corejh Made the requested changes; here's a demo: https://dev-www.chapman.edu/test-section/nick-test/sprint-7-17-2019/html-css-smooth-scroll.aspx#I

nicholasnadel:
> Screen Shot 2019-07-16 at 10 14 12 AM
> 
> Not sure if this is caused by this commit or not but there is an odd empty space at the bottom of the page.
> 
> Second, for my other request i was thinking more along the lines of this
> 
> https://eager.io/blog/communicating-between-javascript-and-css-with-css-variables/
> 
> But maybe that would require more time, so for the meantime in both omninav js and variable.scss add notes to each that changing one you will have to change the other

The footer issue seems to be related to something else; I noticed the same thing months ago on production www.chapman.edu

I've added comments to clarify the SCSS variables and JS function until we have bandwidth for sharing the variables. I guess an intermediate step might be something like the following (where height is set by `getOmniNavHeight()`, but I'll hold off for the time being:

```
  setOmninavHeightScssVariable() {
    $(':target:before').css('height', height)
  }
```

mandythomas:
@nicholasnadel  [posted this to wrong pull a few minutes ago, redoing for 505]: Yesterday I did a pull request (and merge) into Development that included some files which are now showing up in this pull request. My first question is why. Did you do your compare to Development a while back before I'd done those, or today? For the most part it’s showing the same changes I made in various files, however in defaults.scss I don’t see the new “table caption” styles I had added yesterday. They were on line 339 in the file that is now in Development branch but they aren’t showing in this pull request. If I’m reading it right, seems like my change will be missing when you merge.

nicholasnadel:
> @nicholasnadel [posted this to wrong pull a few minutes ago, redoing for 505]: Yesterday I did a pull request (and merge) into Development that included some files which are now showing up in this pull request. My first question is why. Did you do your compare to Development a while back before I'd done those, or today? For the most part it’s showing the same changes I made in various files, however in defaults.scss I don’t see the new “table caption” styles I had added yesterday. They were on line 339 in the file that is now in Development branch but they aren’t showing in this pull request. If I’m reading it right, seems like my change will be missing when you merge.

@mandythomas It initially made this PR 8 days ago, so it likely did not compare against your changes. I've just merged development into this branch (not vice versa!), so your changes should now exist in defaults.scss (https://github.com/chapmanu/cascade-assets/blob/70ce4d298f3002fa93f275a79fd91c184d24eb5c/app/assets/stylesheets/base/defaults.scss)

PR #506, opened 2019-07-15
Misc format and block fixes

mandythomas:
Misc:
For the the OG And Canonical Tags format, I'm fixing an issue with duplicates of the meta tags (I think the previous Developer tried to group the meta tags together with the new ones for twitter but left them in 2 places in the file, plus removed the fully-qualified domain which we need so adding back.

Fix to remove Panther from Panther Alert heading in the homepage alert.

Small update to 2 links in Law's global nav  menu

Updates to depts in dropdown for Faculty Widget in the 2 and 3 Column Data Definitions.

style for table captions.

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/Alert.vtl
+++ b/.cascade-code/Chapman.edu/_cascade/formats/Alert.vtl
@@ -25,7 +25,7 @@
           <path d="M18 26c0 1.105-0.895 2-2 2s-2-0.895-2-2c0-1.105 0.895-2 2-2s2 0.895 2 2z"></path>
           <path d="M16 22c-1.105 0-2-0.895-2-2v-6c0-1.105 0.895-2 2-2s2 0.895 2 2v6c0 1.105-0.895 2-2 2z"></path>
         </svg>
-        <p class="alert-heading">Panther Alert</p>
+        <p class="alert-heading">Alert</p>
     #end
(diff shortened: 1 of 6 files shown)


mandythomas:
@corejh this is not related to that Trello card. I did not update the templates which where the meta tag mentioned in that card reside. I updated a different set of meta tags that are in one of the velocity formats

mandythomas:
@corejh and @cbryant24 I still have pending template changes on Dev-Cascade (swapping out the PRIMARY CONTENT region for DEFAULT). So if Chris is doing that other template change soon we'd need to coordinate or wait til I move my template changes to Prod. I wouldn't want him to simply copy the template from Dev to Prod as is right now. i just haven't got around to finishing that project

cbryant24:
@mandythomas I was probably going to make those changes tomorrow most likely but I can hold off to another day on those if you need more time

mandythomas:
@cbryant24 the meta tag change for the template that you're going to do is (I think) to simply remove the chrome=1 in the 2 and 3 Col modular templates. If you simply edit those on DEV and then make same edit on PROD that would be fine. Just DO NOT copy the version from DEV to PROD as that will bring over the changes I made for PRIMARY vs DEFAULT region names. also fyi I have not made that change to anything in github yet. My chg to template and configs was only in Dev-Cascade so far

PR #509, opened 2019-07-17
Homepage Newsroom Widget Validation

cbryant24:
Story: https://trello.com/c/iWFpSrEB


corejh:
In the future we need need to add creating a unit test as part of jobs like this. Besides scanning the code there is nothing we can do to verify if it's actually working or not

cbryant24:
@corejh I added in a check after the network request has been resolved to verify if it has the correct amount of stories returned from newsroom. Maybe we can use a library like [Joi](https://github.com/hapijs/joi) or [JSON Schema validation](https://github.com/epoberezkin/ajv) to verify the data

mandythomas:
@cbryant24 the Homepage News Stories.vtl in this pull request still shows references to images on News server. I see that you made some mods to the paths to images in this same format in Dev-Cascade. Shouldn't this format in github match?

cbryant24:
good catch @mandythomas it should now match the img urls that are on the cascade format

mandythomas:
@cbryant24 In app/views/widgets/home_page/_newsroom_stories.html  I see paths for the 3 story images that are still going to the News server for that image (both in the IMG tag and the background-image of the DIV). Don't all those need to go to a copy of the images in _files/img/ like you did in some other places ...in case News server is down?

cbryant24:
@mandythomas I updated the urls, I'm pointing them to the images on cascade since looking at the other widgets they have in the views directory that seems to be the normal practice.

PR #515, opened 2019-07-23
Update - make buttons not required on 1 column masthead

nicholasnadel:
story: https://trello.com/c/VeahXzTR

demo:
 - Backend: https://dev-cascade.chapman.edu/entity/open.act?id=05cd7afec0a81e4b18acb60551869e0a&type=page
 - Frontend: https://dev-www.chapman.edu/test-section/nick-test/sprint-7-31-2019/subbrand-masthead-optional-components.aspx


nicholasnadel:
@mandythomas please review when you're able. I tried the approach we'd discussed via email, but opted for this route for content creator usability reasons

mandythomas:
@nicholasnadel the file in github at ..cascade-code/Chapman.edu/_cascade/formats/modular/data-definition.vtl doesn't match up (filename-wise) to any data def we have in Cascade. i realize you probably didn't create this but could we change it as part of this pull request. i am assuming it is a data def for either the 2 Column or 3 Column. The filename in github should reflect the same name. And if there is only one, then there should be two and they should be slightly different. The 3 Column has the Right Column group of fields, and some widgets in Primary Content are excluded. I think there is some cleanup there before you deploy this - essentially a rename and then adding the second data def, by the right name, and with same updates you made to 2 Column

nicholasnadel:
> cascade-code/Chapman.edu/_cascade/formats/modular/data-definition.vtl

I've removed `.cascade-code/Chapman.edu/_cascade/formats/modular/data-definition.vtl` and replaced `app/data_definitions/from_cascade/one_column.xml` with its contents. In the future, it seems like we should restructure Cascade Assets "from_cascade" to better mirror Cascade. 

Please let me know if this change is what you had in mind @mandythomas

mandythomas:
@yes that is what I had in mind. I approve the PR. And, I very much want to get Github to match the folders and filenames in Cascade. some are missing, some are in different paths than Cascade, some should be moved into subfolders in both places for better organization eg. the formats for homepage should all go in /home

mandythomas:
@nicholasnadel Heads up. I compared the 1 Column data def in production Cascade with the one in this pull request and they are slightly different. Do a compare yourself to make sure you aren't undoing someone else's recent updates. It appears that a multiImageContentBlocks widget in Primary Content is live but not in github.

nicholasnadel:
> @nicholasnadel Heads up. I compared the 1 Column data def in production Cascade with the one in this pull request and they are slightly different. Do a compare yourself to make sure you aren't undoing someone else's recent updates. It appears that a multiImageContentBlocks widget in Primary Content is live but not in github.

thanks. will do

PR #518, opened 2019-07-26
Fixes JS error

cbryant24:
Fixes JS error on pages where newsroom homepage widget is not located, prevents dom from being parsed or updating attributes on dom elements that are not present


cbryant24:
@corejh correct you are. Should be removed.

PR #522, opened 2019-08-07
Feature - collapsible accordions toggle all at once

nicholasnadel:
https://trello.com/c/5q2WFgcq
https://dev-www.chapman.edu/test-section/nick-test/sprint-8-13-2019/collapsible-widget-expand-all.aspx


nicholasnadel:
> Looks like when I tab onto it, it has default browser styling. It need's focus element styling consistent with the other links/buttons (dotted red lines)

nice catch @corejh -- Firefox inherited the dotted red but Chrome didn't. I've fixed it at the demo: 
https://dev-www.chapman.edu/test-section/nick-test/sprint-8-13-2019/collapsible-widget-expand-all.aspx

corejh:
@nicholasnadel It should be using margin-top instead of padding so it doesn't add the weird space when it's focused

nicholasnadel:
> @nicholasnadel It should be using margin-top instead of padding so it doesn't add the weird space when it's focused

should be resolved: https://dev-www.chapman.edu/test-section/nick-test/sprint-8-13-2019/collapsible-widget-expand-all.aspx

PR #524, opened 2019-08-23
Feature - fix the h2, h3 (and paragraph) on www to look more hierarchical

nicholasnadel:
https://trello.com/c/zDvo3rYB


nicholasnadel:
> is this an intentional include?
> 
> subprojects/degrees-and-programs/source/static/app/assemble/includes/_css-legacy.hbs
> 
> `<link rel="stylesheet" href="http://localhost:5000/_assets/master.css"/> `

yes; was very annoying to eventually find that it was only pulling master.css from production. I think there should be some indication for whoever has to deal with this next

corejh:
> yes; was very annoying to eventually find that it was only pulling master.css from production. I think there should be some indication for whoever has to deal with this next

Is this the copy of the file that is going into production? if so we need to find a better way to deal with this, otherwise the page will be trying to pull a stylesheet that doesnt exist.

nicholasnadel:
> > yes; was very annoying to eventually find that it was only pulling master.css from production. I think there should be some indication for whoever has to deal with this next
> 
> Is this the copy of the file that is going into production? if so we need to find a better way to deal with this, otherwise the page will be trying to pull a stylesheet that doesnt exist.

Removed for now. I maybe wrongly assumed that it was a handelbars template file that wouldn't end up in production... but I suppose it prob ends up compiled

PR #525, opened 2019-08-23
remove unnecessary  margin-top that was causing issues when multiple …

nicholasnadel:
This is a follow-up to an approved https://github.com/chapmanu/cascade-assets/pull/522. Mandy noticed that the 'expand all' toggle was being pushed below the accordion content. 

demo: https://dev-www.chapman.edu/test-section/nick-test/sprint-8-13-2019/collapsible-widget-expand-all-fix.aspx

### DIFF
--- a/app/assets/stylesheets/widgets/primary_column/collapsibles.scss
+++ b/app/assets/stylesheets/widgets/primary_column/collapsibles.scss
@@ -137,7 +137,6 @@
   cursor: pointer;
   border-bottom: 2px dotted ;
   color: $cu-red;
-  margin-top: 1.5rem;
 
   &:focus {
     outline: unset;


nicholasnadel:
> I only noticed just now, but when I hit expand all, it opens ALL the collapsibles, as opposed to just the section it's attached to. Is this expected behavior?

As a user who's tired of manually clicking each single section link, I can't imagine they'd be more aggravated clicking one final 'expand all'. My understanding is that it's intentional

mandythomas:
@corejh yes "expand all" is supposed to expand all collapsibles at once. You already can do one at a time which is just clicking on the icon. This was to see them all opened, either to print, or use a page FIND tool, or just to view.

PR #535, opened 2019-09-04
New Html Block widget

mandythomas:
new widget to allow embedding a block of html, iframe or javascript into a page in Cascade. Just a text box, not wysiwyg editor. Specifically created for Chris Bryant to put the emergency/rave feed into a page but can be used for other things too. Is for the 2 and 3 Column Modular templates. Allows the code to be entered manually, or to pull in a block stored in /blocks folder in Cascade. either way.


nicholasnadel:
@mandythomas could we limit the visibility/permissions for this to the devs (see Ross' email)?

PR #536, opened 2019-09-09
restructure expand/collapse toggle 

nicholasnadel:
The expand/collapse toggle will now display even if there is no 'intro copy' text

demo: https://dev-www.chapman.edu/test-section/nick-test/sprint-early-september-2019/collapsible-widget-expand-all-fix.aspx

Screen Shot 2019-09-09 at 10 52 37 AM

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/modular/widgets/Collapsibles.vtl
+++ b/.cascade-code/Chapman.edu/_cascade/formats/modular/widgets/Collapsibles.vtl
@@ -10,12 +10,14 @@ $element.getChild('collapsibleRegions').getChild('copy'), true ) )
         <h2 class="sectionHeader">${_EscapeTool.xml($mainheader)}</h2>
     #end 
     #if ($element.getChild('collapsibleRegions').getChild('copy').value != "" or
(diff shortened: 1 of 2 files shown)


nicholasnadel:
> I understand it would be a weird situation and probably not happen, but I think if the collapsible has zero tabs, it will still show the expand all. It probably shouldn't. (I believe this was a problem in the previous version as well, just didn't think of it at the time)

@corejh @mandythomas  for consistency's sake, would it best for these Collapsibles with no children to simply read 'Expand', 'Collapse' (without 'All'), or remove the toggle entirely?

mandythomas:
There will never be a collapsible widget without at least one region. first there'd be no reason to add it and then not fill it in, but if someone did, the header is a required field so they can't submit the page and leave the header empty for the first or any other regions they add to the widget. so just always expect there will be one and show the Expand All ink

nicholasnadel:
> There will never be a collapsible widget without at least one region. first there'd be no reason to add it and then not fill it in, but if someone did, the header is a required field so they can't submit the page and leave the header empty for the first or any other regions they add to the widget. so just always expect there will be one and show the Expand All ink

@corejh are you alright with this and the fca4b62?

PR #540, opened 2019-09-18
Fixes to Dept Contact LC widget

mandythomas:
changes to font and heading styles and how pull in some of the data fields in the Dept Contact widget for the Left Column. For some formats, just bringing them to github for the first time but they already existed in Cascade.

### DIFF
--- a/app/assets/stylesheets/components/left-nav-sub-content.scss
+++ b/app/assets/stylesheets/components/left-nav-sub-content.scss
@@ -109,7 +109,7 @@
           color: $cu-red;
           width:100%;
           padding: 0;
-          font-size: 18px;
+          font-size: 22px;
           margin-bottom: -9px;
         }
       }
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/components/left-nav-sub-content.scss
```diff
@@ -109,7 +109,7 @@
           color: $cu-red;
           width:100%;
           padding: 0;
-          font-size: 18px;
+          font-size: 22px;
```
nicholasnadel: Minor question, but is it worth using rem for new changes?

mandythomas: @nicholasnadel probably but this is just for a little triangle icon and there are a bunch of other pixel settings in this stylesheet that I wasn't planning on changing and testing right now. Just trying to get a quick fix in because the current style and sizes for the Dept Contact widget are wacky. Can we come back to that?

PR #541, opened 2019-09-23
Update the newsletter App.

cbryant24:


mandythomas:
@cbryant24 My fault but you need to fix, sorry: I had modified the link to the Course atalogs link in the Academics dropdown in the omni_nav_v2.vtl file a month+ ago but only in Cascade and didn't update Github. So, this github copy needs the same fix. On line 659 it uses site://Chapman.edu/academics/course-catalogs/index for "Course Catalogs" but should link to https://catalog.chapman.edu/ now. And I guess the _header.html which is a static copy for localhost needs same fix.
Second thing is that we've never put the "form blocks" in Github. They are more like content than programming that supports the templates or site in general. Normally they are not something the Dev team does. You can leave it in this pull request as documentation if you like but none of the others are here nor do I intend to start doing that. We will move all forms to formStack soon anyway.

mandythomas:
@cbryant24 I noticed you hadn't changed the omni-nav-v2.vtl yet for the fix to Catalogs link, so just pinging you about that needing to be done before you deploy

cbryant24:
@mandythomas I made the changes to the course catalog for `omni_nav_v2.vtl` file and removed the marketing form as well since we don't need those in the repo

PR #546, opened 2019-10-08
Accessibility - outdated browser alert

nicholasnadel:
https://trello.com/c/uICFT23I


nicholasnadel:
> Can we get the actual notice itself to be more in line with how the rest of the site looks? Specifically the buttons and add some more margin so it looks a tiny bit better.

updated: https://dev-www.chapman.edu/test-section/nick-test/index-outdated-browser-alert-accessible.aspx/#test-bu

PR #554, opened 2019-10-22
Defect - Collapsible Regions 'Open By Default' Not Working Properly ++

nicholasnadel:
https://trello.com/c/vPoGxydD
https://dev-www.chapman.edu/test-section/nick-test/collapsible-improvements.aspx
https://dev-www.chapman.edu/test-section/nick-test/dmac-collapsible-fix.aspx


mandythomas:
@nicholasnadel Why is there omniNav files mixed in this pull request?

nicholasnadel:
> @nicholasnadel Why is there omniNav files mixed in this pull request?

The Two Column Template in the Modular Controller was still using an older Omninav design from a  previous dev team. I had to add the current production v2 design as a partial in order to test offsetting its height when toggling the accordions and scrolling to the respective accordion section

app/controllers/content_types/modular_controller.rb
```diff
@@ -247,7 +247,7 @@ def two_column
         "SOCIAL ACCOUNTS" => "TODO: _cascade/formats/level/social_accounts",
 
         # TODO: convert these to cascade_format action.
```
nicholasnadel: The Two Column Template in the Modular Controller was still using an older Omninav design from a previous dev team. I had to add the current production v2 design as a partial in order to test offsetting its height when toggling the accordions and scrolling to the respective accordion section

mandythomas: ok makes sense

PR #566, opened 2019-11-01
remove link to Graduate Research in omniNav

mandythomas:
remove link that's currently in omniNav and hamburger's Main Menu in the Research section. It's the Graduate Research link. Per SMC that site is being retired.


nicholasnadel:
@mandythomas  please review https://github.com/chapmanu/cascade-assets/pull/565, would like to include it if you will be pushing this today

PR #569, opened 2019-11-05
replace duplicate <footer> with semantic <div>

nicholasnadel:
https://trello.com/c/tnyJ6jCq


.cascade-code/Chapman.edu/_cascade/templates/Modular/1_column.html
```diff
@@ -1,56 +1,60 @@
...
+		<meta content="IE=edge" http-equiv="X-UA-Compatible"/>
+		<title>
+			<system-page-title/> | Chapman University
+		</title>
+		<system-page-meta-keywords/>
+		<system-page-meta-description/>
+		<system-page-meta-author/>
+		<system-region name="OG_TAGS"/>
+		<system-region name="META VIEWPORT"/>
+		<!-- Jquery, Typkit, Common Assets -->
```
cbryant24: @nicholasnadel not a big deal just a type, supposed to be `Typekit`

nicholasnadel: typos are cool now https://twitter.com/ow/status/1191799104919216130

fixed below

mandythomas:
@nicholasnadel Is .cascade-code/Chapman.edu/_cascade/templates/Modular/default.html in there for a reason? Is it new? I don't think we've ever used that for anything. I don't see that in Cascade itself.

mandythomas:
@nicholasnadel hard to compare all the templates as they show the whole thing changed. Is the gist just that the indenting is fixed and you've modified the <footer> tag to be a <div>. Anything else to look out for?

nicholasnadel:
> @nicholasnadel hard to compare all the templates as they show the whole thing changed. Is the gist just that the indenting is fixed and you've modified the
> tag to be a
> . Anything else to look out for?

I haven't edited anything besides wrapping the footer in the `<div class="footer__container">` as we discussed. I copied and pasted from Cascade's edit view, with their 'Format Code' applied

Screen Shot 2019-11-06 at 10 32 18 AM

nicholasnadel:
@mandythomas it may be clearer if you hide the whitespace changes 
Screen Shot 2019-11-06 at 10 33 37 AM

mandythomas:
@nicholasnadel yeah I think we should remove Default in both github and dev-cascade. It appears from the version history to be something you created in February, then I edited it in June when I modified all the templates to go from the PRIMARY CONTENT region to DEFAULT. I think it was just something we had for practice. It doesn't exist on production Cascade

nicholasnadel:
> @nicholasnadel yeah I think we should remove Default in both github and dev-cascade. It appears from the version history to be something you created in February, then I edited it in June when I modified all the templates to go from the PRIMARY CONTENT region to DEFAULT. I think it was just something we had for practice. It doesn't exist on production Cascade

Alright @mandythomas I've removed from version control and on dev-cascade

PR #572, opened 2019-11-21
Defect - homepage bigs stats widget

nicholasnadel:
https://trello.com/c/SlVv1TZX

I wanted to get your opinions to see if this is a reasonable/valid solution. If not, I'll remove the counter animation and add a Trello card until we have more bandwidth.

The issue seems to stem from the JavaScript occassionally changing the actual default `original_num` value.

I've added a `data-count` attribute to the velocity format, which takes the initial value of `$factNumber` and some JS to check if that attribute exists (since other pages seem to rely on the `animateSingleNumber` function and might not have `data-count`.  

for demo purposes, the velocity format's first fact text value is '666':
https://dev-www.chapman.edu/test-section/nick-test/homepage-big-stats-2.aspx/#undergraduateAdmission

update: added opacity on initial load after discussing with Mandy
https://dev-www.chapman.edu/test-section/nick-test/homepage-big-stats-3.aspx/#undergraduateAdmission

### DIFF
--- a/.cascade-code/Chapman.edu/_cascade/formats/Homepage UG Admission Section
+++ b/.cascade-code/Chapman.edu/_cascade/formats/Homepage UG Admission Section
@@ -43,23 +43,23 @@
                     #set ($factLabel = $fact.getChild('Label').value )
                     #if ($foreach.count == 1)
                         #set ($position = "120px" )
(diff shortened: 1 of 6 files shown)


corejh:
For this particular fix I can't reliably break the thing so I'll leave it to @cbryant24 and @mandythomas

PR #575, opened 2019-11-26
Fix collapsible widget

cbryant24:
Story: https://trello.com/c/u5qINseg


cbryant24:
@mandythomas good catch didn't notice that debugger. It should be removed now.

PR #576, opened 2019-11-26
WAI-ARIA image is missing alternative text

cbryant24:
Story: https://trello.com/c/zTT9Nmyr


cbryant24:
@corejh the icons should now all have `aria-hidden="true"` and `role="presentation"`

cbryant24:
@mandythomas I added the `<desc>  and <title>` fields to fix the alt text accessibility issue but changing `role="presentation"` and `aria-hidden="true"` fixed it without the need of adding `<desc> and <title>` fixed the accessibility issue so I removed those elements to match the other svg elements.

cbryant24:
@mandythomas removing those also fixed the typo issue as well.

PR #577, opened 2019-12-03
Redesign - Callout Widget

nicholasnadel:
https://trello.com/c/OuBSmbnO

PR #579, opened 2019-12-05
Callout - redesign fix conflicts

nicholasnadel:
https://trello.com/c/OuBSmbnO
Follow-up to #577

This resolves some merge conflicts with development and style conflicts with the [department contact widget](https://dev-cascade.chapman.edu/entity/open.act?id=ea5dd1c6c04d744c7b41aa25c7cdb996&type=page ).

New features worth noting:

- includes CSS variables in variables.scss. This allows quick referencing of our color palette directly from the browser inspector (outside of Rails) using `color: var(--variableName)`:

- Screen Shot 2019-12-05 at 9 33 52 AM

- vastly improved font rendering in Firefox on OS X with `-moz-osx-font-smoothing: grayscale;` in defaults.scss

demos:
- 2 col: https://dev-cascade.chapman.edu/entity/open.act?id=a4546928c0a81e4b22b523b8a8e91bde&type=page
- 3 col: https://dev-cascade.chapman.edu/entity/open.act?id=a4e5c0f5c0a81e4b22b523b8b2dd1841&type=page


nicholasnadel:
> @nicholasnadel base/defaults.scss has commented-out lines. I think we are supposed to remove those.
I've removed the unnecessary comments from this branch.

> Also, wondered why you're changing home.scss for a project about left nav callouts?
[The new designs call to remove the grey background and window pattern across the site](https://trello.com/c/cF4MaTQz). That line in home.scss was overriding the new white background in defaults.scss: https://recordit.co/zR11uxz7nH

> Last, in both the two and three-column data definitions, the new color-value field is inside a color-wrapper group but there's only that one field in the group. I'd delete the group, not needed. 

I've removed this group.

> Also as a best practice, once within a particular widget's node I don't think any of the groups within it should be set to collapsed by default. Leave open so easily seen. 

Okay.

> In same 2 data defs, the Text field has [default] values in the fields. Leave them blank and put suggestions for content in the Help Text instead.

Replaced that existing placeholder text with help-text

nicholasnadel:
> im apprehensive on having all the colors added to root like it is, but im unfamiliar with any cons so tentatively approving

https://lisilinhart.info/posts/css-variables-performance 

If I'm interpreting correctly, sounds like performance is mainly a concern if there are multiple children? 

`So the more children a parent element has using this variable, the more expensive setting a CSS variable on this element gets.`

If this is a bad idea, I'll revert, but I do think it would be useful when eyeballing changes in the inspector outside of the Rails app

PR #580, opened 2019-12-09
Multi image content block - defect -- padding

nicholasnadel:
Multi Image Block Fix (for PJ https://trello.com/c/PQVQHddA)
before (‘Test Block for Development’ near footer): https://wwwtest.chapman.edu/law/index-new.aspx
after: https://dev-www.chapman.edu/test-section/nick-test/multi-image-content-block__padding-etc.aspx


.cascade-code/Chapman.edu/_cascade/formats/modular/1 Column Primary Content.vtl
```diff
@@ -28,79 +27,83 @@
...
-    
+
     #if ($widgetType == 'Contact Footer')
         #outputContactFooter($widget)
     #end
-    
+
     #if ($widgetType == 'Multi Image Content Blocks')
-        #outputMultiImageContentBlocks($widget)
+        #set ($displayToggle =
```
corejh: This is interesting. It might be a nice idea to implement this as a standard option across the board if it makes sense

PR #581, opened 2019-12-10
Web Services - 'Pull' Rake Task

nicholasnadel:
This is from a story a few months ago. 
It keeps coming in handy to download multiple files in a directory (in my case the new OmniNav and its 10s of partials)

Interesting precursor to this from 2017: https://trello.com/c/oBBHmLj3/471-spike-test-cascade-web-services-api-using-a-command-line-http-client


nicholasnadel:
see https://github.com/chapmanu/cascade-assets/pull/682

PR #656, opened 2020-05-11
BUG - Left nav on IE is not showing the back button

cbryant24:
story: https://chapman.my.workfront.com/task/view?ID=5eb1e1500497d6e783163db03ac0893d


app/assets/stylesheets/base/mixins.scss
```diff
@@ -491,11 +491,53 @@ ORDER: Base + typography > general layout + grid > page layout > components
...
+    // our css4 variable output
+    $var-output: var(--#{$map-name}-#{$var});
+  }
+  // if show fallback is not overridden to be null
+  @if $show-fall {
+    #{$prop}: $var-fall;
+  }
+  // css4 variable output
+  #{$prop}: $var-output;
 }
```
cbryant24, line 534: This mixin is used to allow for css variables and provide a fallback for Internet Explorer 10 & 11 which can't use css variables. It works by passing in the css property and corresponding variable. Sass allows for lists so you can group varibales together using key value pairs. This mixin parses thoses lists and creates the css property using the Sass varibale and creates a css variable as well adding the css property to the front of the variable see the screenshot for the otuput of this mixin

cbryant24: Screenshot 2020-05-11 15 28 00

app/assets/stylesheets/base/mixins.scss
```diff
@@ -603,25 +645,25 @@ ORDER: Base + typography > general layout + grid > page layout > components
   ul li {
     display: flex;
     justify-content: space-between;
-    border-bottom: 1px solid v(cu-light-grey);
+    @include v(border-bottom, solid-grey);
```
cbryant24, line 649: Another example of how the mixin works and the css variable it creates using the css property combined with the Sass variable

cbryant24: Screenshot 2020-05-11 15 31 39

nicholasnadel: This is dope; I have two concerns:

1) If we don't leave the current `var(--cu-red)` in place, we can't use them in the browser inspector, which I use often to eyeball colors

Screen Shot 2020-05-11 at 3 38 43 PM

2) Dependent on 1) (if we don't leave the vars in place), we should do a search all and replace the few straggler CSS vars using that syntax. Did that [here](https://github.com/chapmanu/cascade-assets/tree/ie-fix__nick) if you want to merge.

I'd really prefer leaving the CSS vars in 1) if you all are cool with it. Maybe we should move all the colors back to their own SCSS file for organization

nicholasnadel: Lastly, would it be less work to just use a polyfilly since we're only worried about IE?
https://github.com/nuxodin/ie11CustomProperties 
https://github.com/polyfills/variables

cbryant24:
@nicholasnadel @corejh I followed [this tutorial](https://codepen.io/jakealbaugh/post/css4-variables-with-fallbacks-using-sass) For using CSS variables with a fallback. This is essentially so we can use CSS variables with IE10 & 11. I think it allows us to organize or variables better as well grouping them into lists. It builds a mixin that creates a fallback property when using css variables. If you have any questions let me know.

cbryant24:
@nicholasnadel sorry I think i removed your comment not sure how. But couldn't you search for color using the css variable but searching `--color-cu-red` also we can remove the `color` portion of the variable because it just prefixes it with the css property in this part of the `@each` Sass loop 
```css
      } @else {
        // do the definition
        #{--$var-map-name}-#{$var}: $val;
      }
```

cbryant24:
@nicholasnadel 
Screenshot 2020-05-11 16 08 34

cbryant24:
@nicholasnadel @corejh otherwise according to [this stackoverflow](https://stackoverflow.com/questions/45955538/workaround-for-css-variables-in-ie) this would be the other option which is reflected in this previous commit before this version https://github.com/chapmanu/cascade-assets/pull/656/commits/4f3e4d4e3fd82e460ec7b477d3b7180b7ada28ac

nicholasnadel:
> @nicholasnadel sorry I think i removed your comment not sure how. But couldn't you search for color using the css variable but searching `--color-cu-red` also we can remove the `color` portion of the variable because it just prefixes it with the css property in this part of the `@each` Sass loop
> 
> ```css
>       } @else {
>         // do the definition
>         #{--$var-map-name}-#{$var}: $val;
>       }
> ```

I missed the --color prefix when testing. So cool. 

The only other concern was any straggler properties using the `var(--color)` syntax instead of `v()`: https://github.com/chapmanu/cascade-assets/tree/ie-fix__nick

cbryant24:
@nicholasnadel take a look at [the homepage on dev](dev-www.chapman.edu) I updated `app/assets/stylesheets/widgets/_ad_landing_page/homepage_masthead.scss` from css variables to Sass variables some I had to hardcode though

nicholasnadel:
> @nicholasnadel take a look at [dev homepage masthead](https://dev-www.chapman.edu/) I updated `app/assets/stylesheets/widgets/_ad_landing_page/homepage_masthead.scss` from css variables to Sass variables some I had to hardcode though

Well done, looks good to me

PR #689, opened 2020-08-15
Adding Webpack Asset Files To Cascade Rake Task

cbryant24:
Needed to add the webpacker output CSS and JS files to the rake task `edit_cascade_assets`


cbryant24:
> Looks good to me.
> 
> Side note, we can add `_backup/*` to `.gitignore` to avoid accidentally committing any `.bak` files -- the caveat is they'd obviously be wiped out on `git reset --hard `
> 
> Up to yall

@nicholasnadel added `/_backup` to gitignore

PR #718, opened 2020-10-21
Navigation hiding links with anchor tags

cbryant24:
Story: https://chapman.my.workfront.com/task/5f372127004d5c4797c1112194927cb2/overview


nicholasnadel (approved):

corejh (changes requested):
This isn't a hotfix. This is a permanent fix that will stay on the site. Maybe just renaming the namespace/folder is fine. 

I can't think of a way to do it off the top of my head, but if there's a way to have the offset constants reference a uninav height variable, that would be ideal.. not a dealbreaker if there isn't a quick solution for that. Just need to make some notes this needs to update if the other ever does

cbryant24:
@corejh I updated the file structure let me know if you think the `considerations` directory works. Also I tested it on this page http://dev-www.chapman.edu/a-z/index.aspx and works with the new fix.

PR #728, opened 2021-03-03
Custom navigation

nicholasnadel:


corejh (changes requested):
there are a lot of `console.log` scattered around. can we comment or remove them from production

nicholasnadel:
> there are a lot of `console.log` scattered around. can we comment or remove them from production

@corejh these logs should be removed -- could you re-review when convenient?

PR #789, opened 2023-03-16
Added issues with M1 mac and 'npm install' to README

kitokazu:
As I was following the instructions on how to get my environment setup, I ran into a few issues with `bundle install` and `npm install`. After looking into it, the bundle install issue may have been a result of using an **M1 mac**. 

Error with `bundle install`:

`There was an error parsing 'Gemfile': cannot load such file -- em/pure_ruby. Bundler cannot continue`

Here is the [resource I found that helped](https://github.com/oneclick/rubyinstaller2/issues/96#issuecomment-432629209). The main command that helped was:
```
gem install eventmachine --platform ruby
```

For `npm install` I had to make sure that my computer was running the version of node that was compatible with the project. [Link to source that helped](https://stackoverflow.com/questions/7718313/how-to-change-to-an-older-version-of-node-js).

I essentially used [n](https://github.com/tj/n) which is a Node version manager that can be installed via npm.

```
npm install -g n   # Install n globally
n <version>         # Install and use <version>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,17 @@ Once you’ve set the default version of the Ruby run “bundle install” again
 bundle install
 ```
 
+#### Issues with M1 Macs: 
+This may be an issue with Macbook that are running the M1 chip, but if you are getting this error when running "bundle install":
+```
+There was an arror parsing 'Gemfile': cannot load such file -- em/pure_ruby. Bundler cannot continue
+```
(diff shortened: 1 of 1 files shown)


d-ahn10 (approved):
Looks good~ Thanks Kai!
Good to see this working on M1 Macs too

